PR #2, opened 2017-07-05
Research into the feasibility & desirability of this

cap10morgan:
Since we decided to move forward with this, it's time we merged this into master. The tests are all on another branch that I will PR right after this merges.

This should contain every HTTP API that user-facing TurboVote needs to work. I have run it in a krakenstein without the other -http-api components and it worked like a champ.

[Pivotal card](https://www.pivotaltracker.com/story/show/145946831)


src/turbovote_http_api/three_scale.clj
```diff
@@ -0,0 +1,40 @@
+(ns turbovote-http-api.three-scale
+  (:require [turbovote.resource-config :refer [config]]
+            [clojure.tools.logging :as log])
+  (:import (threescale.v3.api ParameterMap AuthorizeResponse ServerError)
+           (threescale.v3.api.impl ServiceApiDriver)))
+
+(def api-client (ServiceApiDriver/createApi))
```
zonotope: since 3scale is an external service that we're reaching out to, i think we should put all of our interactions with it behind a boundary protocol, in a different namespace, and implement that protocol here. i'm thinking something like this:

in a new ns called `turbovote-http-api.boundary.authorization`:
```clojure
(ns turbovote-http-api.boundary.authorization)

(defprotocol RequestAuthorizer
  (authorize [authorizer request] 
    "Authorize `request` with the pre-configured `authorizer`."))
```

then, in this ns we could do something like:
```clojure
(ns turbovote-http-api.three-scale
  (:require
   ...
   [turbovote-http-api.boundary.authorization :as authorization]
   ...)

(defrecord ThreeScaleAuthorizer [service-token service-id three-scale-api-client]
  authorization/RequestAuthorizer
  (authorize [_ {:keys [headers] :as request]
    (let [user-key (get-the-user-key-from-the-request)
          usage (generate-the-usage-param)
          params (generate-the-params)]
       (do-the-other-stuff-to-authorize-the-request))))

(defn three-scale-authorizer [service-token service-id]
  (let [three-scale-api-client (ServiceApiDriver/createApi)]
    (->ThreeScaleAuthorizer service-token service-id three-scale-api-client)))

;; and then you can use the `three-scale-authorizer` fn to initialize a 3-scale
;; authorization client that will work with the `RequestAuthorizer` protocol.
```

writing to protocols like this is definitely a different style, but i think it helps keep things tidy by separating our business logic from the intricacies of any particular external api we might be depending on. 

this will make things easier to test because we can just make a dummy record that implements the same protocol, but just says that all requests are authorized (or that no request is authorized, depending on what we want to test), and we wouldn't have to make any dangerous calls to `with-redefs` to do it.

it will also make it much easier to switch to a new authorization api later if we need to because all the api-specific code and logic is isolated. it would just be a matter of defining a new record that implements our protocol, and initializing that record for the api client. 

the code that's here now works, and connecting to 3scale through a protocol might be a substantial architectural change, so if you don't think the advantages i've listed are worth it, or if you think those changes should come in another pr, then i'm fine with that.

src/turbovote_http_api/handlers/elections.clj
```diff
@@ -0,0 +1,71 @@
...
+  (async/close! search-upcoming-channel)
+  (async/close! read-all-upcoming-channel)
+  (async/close! search-create-electorate-channel))
+
+(def authorize-api-request
+  (pedestal/interceptor
+   {:enter
+    (fn [{:keys [request] :as ctx}]
+      (let [authorization (ts/authorize-request request)]
+        (log/debug "3scale API authorization response:" (pr-str authorization))
```
zonotope: minor, but i think this logging statement should be moved to the three-scale ns.

src/turbovote_http_api/handlers/elections.clj
```diff
@@ -0,0 +1,71 @@
...
+
+(defn close-all-channels! []
+  (async/close! search-upcoming-channel)
+  (async/close! read-all-upcoming-channel)
+  (async/close! search-create-electorate-channel))
+
+(def authorize-api-request
+  (pedestal/interceptor
+   {:enter
+    (fn [{:keys [request] :as ctx}]
```
zonotope: minor, but i think this fn could use a test.

cap10morgan:
@zonotope Thanks for the review! The tests are all coming in a subsequent PR: #3.

PR #3, opened 2017-07-05
Copy tests from -http-api components

cap10morgan:
~~Once #2 is merged,~~ we should review and merge this too. It brings in the tests from the old separate components, cleans them up, makes them more consistent, and gets them all passing in here.

I also implemented some of the suggestions from #2 in here, such as using a protocol to abstract the 3scale API a little better. That eliminated the use of `with-redefs` in the tests, so that's nice. I also discovered that `io.pedestal.test/response-for` now supports async interceptors (the fact that it didn't before is why we weren't using it). So that also eliminated the need for the test fixture spinning up a real HTTP server. 🎉 

[Pivotal card](https://www.pivotaltracker.com/story/show/147852887)

The http-api tests all pass. This was also tested w/ acceptance tests and a krakenstein signup. I also tested using a 3scale API key to access all elections. Everything worked great.


src/turbovote_http_api/boundary/authorization.clj
```diff
@@ -0,0 +1,54 @@
...
+  (authorize [authorizer api-key request]
+    "Authorize `request` with the pre-configured `authorizer`.")
+  (authorized? [authorizer api-response]
+    "Returns true if `api-response` represents an authorized response, false
+    otherwise.")
+  (failure-reason [authorizer api-response]
+    "Returns a human-readable explanation of why auth failed given a
+    failure `api-response`."))
+
+(defn get-api-key
```
zonotope: i think this in a 3scale implementation detail leaking into this ns. i could imagine a request authorizer that doesn't need an api key in the headers.

src/turbovote_http_api/boundary/authorization.clj
```diff
@@ -0,0 +1,54 @@
+(ns turbovote-http-api.boundary.authorization
+  (:require [clojure.tools.logging :as log]))
+
+(defprotocol RequestAuthorizer
+  (authorize [authorizer api-key request]
```
zonotope: maybe we should be thinking at a higher level for the authorization api. i think boundary protocols should codify what the rest of the application business logic expects to be able to do with it, so it might be better if we thought at the level of the `authorize-request` fn defined below because that's the only thing the rest of the app cares about. plus, your docstring for that fn says as much.

you might want to consider making the `RequestAuthorizer` protocol only include the `authorize-request` method, and move the implementation we have here to the `turbovote-http-api.authorizer.three-scale` ns.

src/turbovote_http_api/boundary/authorization.clj
```diff
@@ -0,0 +1,54 @@
...
+  Authorization: apikey this-is-the-actual-api-key
+
+  The header name and the apikey token are both case-insensitive."
+  [headers]
+  (some-> headers
+          (get "authorization")
+          (->> (re-matches #"(?i)\Aapikey\s+(.+)\z"))
+          last))
+
+(defn authorize-request
```
zonotope: as i wrote above, i think this should probably move to the 3scale implementation ns, and leaving `authorize-request` to just be a protocol method declaration here.

src/turbovote_http_api/handlers/elections.clj
```diff
@@ -8,9 +8,14 @@
    [io.pedestal.interceptor :as pedestal]
    [turbovote.resource-config :refer [config]]
    [turbovote-http-api.handlers.utils :as utils]
-   [turbovote-http-api.three-scale :as ts])
+   [turbovote-http-api.boundary.authorization :as authorization]
+   [turbovote-http-api.authorizer.three-scale :as three-scale])
   (:import (java.util UUID)))
 
+(def ^:dynamic *authorizer*
```
zonotope: could we use `delay` here?

zonotope: ahh, i realize now that this makes writing the tests easier. we probably don't really care about concurrent access because the state is the same regardless, and the performance impact is probably negligible. 

please disregard the parent.

src/turbovote_http_api/handlers/elections.clj
```diff
@@ -22,21 +27,28 @@
 
 (def authorize-api-request
   (pedestal/interceptor
-   {:enter
+   {:name ::authorize-api-request
```
ddellacosta: Can you tell me about the namespaced keyword `:name`s in these interceptors?

test/turbovote_http_api/user_service_test.clj
```diff
@@ -0,0 +1,217 @@
...
+                            root-url
+                            :body (pr-str user-data)
+                            :headers {"Content-Type" "application/edn"})
+                  create-data (edn/read-string (:body response))]
+                (is (= 201 (:status response)))
+                (is (= user-with-id create-data)))))
+    (testing "valid Transit+JSON POST to / returns 201 on success"
+        (let [user-data {:first-name "Troy"
+                         :last-name "Tulowitski"
+                         :email "troy@rockies.mlb.com"}]
```
ddellacosta: We should make up a drinking game about who wrote what test/comment/etc. based on various information like this, where you're not allowed to look at `git blame` or `log` or whatever, of course

src/turbovote_http_api/boundary/authorization.clj
cap10morgan: The problem I ran into that led me to this approach was sharing the response handler code across implementors of the protocol. I found that I was just testing the dummy authorizer before (vs. any code that would get used in production).

But I'm sure there's more than one way to skin that particular cat.

src/turbovote_http_api/boundary/authorization.clj
cap10morgan: Hmm... I don't think I agree. There will very likely always be API keys involved here. And putting it into the header like this is 100% our doing. It has nothing to do with 3scale nor would it be affected if we replaced 3scale with a different API authorization backend (external or internal).

src/turbovote_http_api/boundary/authorization.clj
cap10morgan: I _really_ want to keep all of the response-constructing code common across both authorizers. I'll revisit the way in which I'm doing that, though.

cap10morgan: And actually, as I take another look at it, I'm reminded that I also want to keep all of the rest of the code in `authorize-request` common across all authorizers. It isn't 3scale-specific. A better way to achieve that isn't jumping out at me. Want to discuss in a screen share, @zonotope?

src/turbovote_http_api/handlers/elections.clj
cap10morgan: Sure. We should have been namespacing them all along (they are intended to globally identify the interceptor), but it wasn't a big deal when we had lots of small http-api components that each dealt with a very limited domain. But now that they are all consolidated, you could pretty easily imagine more than interceptor sharing the same local name (`authorize-api-request` is as good a hypothetical example as any) but in a different namespace. When these namespaced names show up in logs and error messages, it's very easy to trace them back to where they live in the code.

ddellacosta: Cool, thanks for the explanation!

src/turbovote_http_api/boundary/authorization.clj
zonotope: i get what you mean about keeping the code common across the different authorizers. we definitely should not have to re-invent the wheel here. 

perhaps you could have some formatting functions that you could call from the specific authorizers. looking at the `authorize-request` implementation, you could do something like this:

```clojure
(defn authorized [response]
  {::status ::authorized
   ::auth-response response})

(defn unauthorized [response message]
  {::status ::not-authorized
   ::auth-response response
   ::message message})

(defn auth-error [error-data message]
  {::status ::error
   ::error-data error-data
   ::message message})
```

this will only make sure that the return values are all the same shape. you'd still have to repeat all the control flow logic and logging in every implementation, so this doesn't help you with that. you *could* make a helper function that implements that control flow, but that's just going around in circles at that point.

src/turbovote_http_api/boundary/authorization.clj
zonotope: i see what you mean now. thanks for the explanation. i misunderstood where this api-key was coming from before. since it's something we add to the request somewhere else, i'd argue that it is part of our business logic now and not 3scale specific.

src/turbovote_http_api/boundary/authorization.clj
zonotope: i hear you about sharing non-authorizer-specific code, and i have some less than ideal suggestions for that below. 

as far as testing goes, i usually test each specific protocol implementation once to make sure it's doing the right thing. then i test the rest of the business logic code that uses the protocols with the dummy. that way everything is still tested, but i don't have to worry about making external api calls all over my test and only focus on those calls where i'm actually testing them specifically.

cap10morgan:
Thanks @zonotope and @ddellacosta!

PR #4, opened 2017-08-21
add endpoint to search election authorities

zonotope:
## Purpose
This is to expose a way for the `turbovote-web` client to find the clerk's address and phone number for the special New Hampshire voting instructions.

## Open Question
This endpoint is completely open right now, but should requests to the new endpoint to search election authorities be authorized somehow?


cap10morgan:
I _think_ the reason there was no endpoint for directly looking up authorities is that all such data should come from elections (via `election-works`) or voter registration rules (via `voter-registration-works`), not authorities. This allows for specific elections to override the default rules in certain cases (those default rules do come from election-authority-works originally, though). Does that not work in this case for some reason, though?

cap10morgan:
To follow up on my first comment: If this is _just_ about voter registration, then you can ignore the elections aspect altogether and just focus on the `/voter-registration/registration-methods/nh` endpoint (that already exists). Currently it just returns `{:by-mail {:acceptable-forms #{{:name :nvrf}}}}` which is incomplete and/or inaccurate perhaps? But adding to or changing what that returns such that a) it is a useful response for a non-TV-specific API whose job is to answer the question (in data): "What are my options for registering to vote in state X?" and b) consuming that response in turbovote-web to implement the TV feature in question. That does seem like a viable path here? I'm happy to dig in deeper if it would be helpful.

zonotope:
@cap10morgan this is in support of [this card](https://www.pivotaltracker.com/story/show/112446043) (which i should have linked to begin with; sorry about that). 

i need to display the election clerk's office name, email, and phone number to the user just after the "where do you want to register" page if they live in new hampshire. the only place in the schemas across all the services that i found this info was in e-a-w, but i'm happy to move it to another place and avoid opening up this endpoint if we don't want to expose e-a-w to the world.

i don't think election-works will be a good fit as y'all mentioned in the other pr because there might not be an election at that point when the user is signing up. 

this is dealing with voter registration methods, and part of this work is adding a new `:call-clerk` registration method too, so i think the info i need in makes sense the context of the `/voter-registration/registration-methods/...` hierarchy. i can add the `:call-clerk` method to the endpoint that's already there, but i think i will also need a new one to get the name, address, and phone number of the clerk's office the user will need to contact.

what do you think of me adding a new handler to `voter-registration-works` that takes in a list of ocd-ids and returns the contact info for the clerk's offices matching those ocd-ids, if there is any contact info available. then, i could expose an endpoint here like `/voter-registration/registration-methods/nh/election-clerks` that listens for `GET`s with a list of ocd-ids as the query string. does that sound like it could work?

cap10morgan:
@zonotope Have we decided we won't need this after all?

zonotope:
yes. this is no longer necessary, so closing.

PR #5, opened 2017-10-02
Upgrade bifrost to 0.2.0 and use fns in interceptors

tie-rack:
The real benefit will come later when we need asynchronous interceptors that don't map directly to a works component's handler.

Pivotal story: [150491144](https://www.pivotaltracker.com/story/show/150491144)


src/turbovote_http_api/handlers/vote_by_mail.clj
```diff
@@ -2,16 +2,19 @@
   (:require
    [bifrost.core :as bifrost]
    [clojure.core.async :as async]
+   [kehaar.wire-up :as wire-up]
    [turbovote-http-api.handlers.utils :as utils]))
 
 (defonce apply-channel (async/chan 1000))
 
+(def apply-fn (wire-up/async->fn apply-channel))
```
cap10morgan: Why `apply-fn` here but `create` and `update` elsewhere? Just curious.

tie-rack: I caught this in my overview before I pushed these changes up. But I couldn't bring myself to make the change to match the others. The name `apply` feels different. Sacrosanct.

I'll make the change if I can add a comment apologizing to McCarthy.

cap10morgan: Damn, that's such a good argument. I kinda had a similar feeling reading over it and was secretly hoping the reason would be along these lines. You can't just _reject_ `apply` after all.

I am fine with whatever you decide here. But the comment apologizing to McCarthy will make me happy every time I see it. FWTW.

tie-rack: I can't do it.

tie-rack:
This is worth glancing at again @cap10morgan with the new endpoints covered since the last push (and then we can merge if you like it).

PR #6, opened 2017-10-05
Adds unsubscribe endpoint

marlabrizel:
Adds an unsubscribe endpoint and interceptor on top of the `:subscription/delete` endpoint to validate a subscription delete request before passing along. Prevents unauthorized delete from hitting the system

Related PRs:
https://github.com/democracyworks/turbovote-web/pull/230
https://github.com/democracyworks/notification-works/pull/43
https://github.com/democracyworks/notification-templates/pull/45

[Pivotal card](https://www.pivotaltracker.com/story/show/150280176)


marlabrizel:
@ddellacosta good call, I'll add those

src/turbovote_http_api/handlers/utils.clj
```diff
@@ -34,4 +36,14 @@
...
+
+(defn validate-token
+  "Validates a supplied token to ensure that a given request is legitimate
+   prior to executing the specified api call. Returns an error if token cannot
+   be validated."
+  [token]
+  (try
+    (jwt/unsign token (config [:jwt :secret]))
+    (catch
+      Exception e (ex-data e))))
```
zonotope: a few things. first, not all exceptions are instances of `ex-info`, and only instances of `ex-info` return anything when you call `ex-data` on it. if you call `ex-data` on a java exception you get `nil`. 

also, you should also add a status to this payload. something like:

```clojure
(catch Exception e
  {::status error, ::exception e})
```

so you can dispatch on the status in code that uses this function. also, note the namespace qualified keyword that i used for status (`::status` instead of `:status`). this is to make sure your dispatch code doesn't get confused by successfully unsigning a token that happened to have a `:status` key.

src/turbovote_http_api/handlers/election_notifications.clj
```diff
@@ -21,6 +23,21 @@
...
     (async/close! c)))
 
+(def validate-token-interceptor
+  (interceptor
+    {:enter
+     (fn [ctx]
+       (let [unsigned-token (utils/validate-token
+                              (get-in ctx [:request :path-params :user-token]))
+             user-id (java.util.UUID/fromString (:user-id unsigned-token))]
+         (if (not= {:type :validation :cause :signature}
```
zonotope: there's a potential security hole here. if the `utils/validate-token` function throws an exception for any other reason, we'd still treat the request as authorized. i suggested later on that you add a status to the map you send back from `utils/validate-token` in case you catch an exception so you can better tell when an error occurred. 

here, you also don't want to check that the return value matches exactly because there are many more error cases than this one. instead, you should do something like:

```clojure
(if (= (::utils/status unsigned-token) :error))
  ...)
```

PR #7, opened 2017-10-10
Add Twilio callback endpoint

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/151665831)

Intended to work with zendesk-works, so maybe don't merge until that's ready (though I don't think it will hurt anything if we do).

We should move Twilio creds in Consul from being nested under a top-level `sms-works` key to just being under a top-level `twilio` key. Then we'll already have the account-sid value we need there.

I'll put up a companion PR for sms-works to work with that and push another commit here as well. Then we can just have the creds in both locations in Consul for a bit until the new sms-works is rolled out everywhere.


cap10morgan:
Verified that this puts the SMS on the `zendesk-works.sms.callback` RMQ channel now.

cap10morgan:
Put up an [sms-works PR](https://github.com/democracyworks/sms-works/pull/15) to access Twilio creds from a root `twilio` folder in Consul.

tie-rack:
Approved as part of #8. Closing this PR to let that one take care of it.

PR #8, opened 2017-10-13
Zendesk callback

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/151666436)

Based on the `twilio-callback` branch, but I suppose it doesn't need to be. Let me know if you'd prefer a master-based version instead.

Anyway, this adds a `zendesk` endpoint for us to point their ticket comment callback at so that comments on SMS-created tickets can be forwarded to zendesk-works for forwarding back to the original texter.


cap10morgan:
NB: I haven't added the ZenDesk creds to any Consul clusters yet because I don't think we know what they'll be yet.

PR #10, opened 2017-12-20
Fix endpoint association

cap10morgan:
Fixes [this bug](https://www.pivotaltracker.com/story/show/153818417). Fix verified in krakenstein.

This will be important as we add more API endpoints that 3scale will handle authn/z for.

Also updated Clojure, added a project description, and fixed a shutdown hook bug I noticed. If any of these seem inappropriate as targets of opportunity, let me know!


src/turbovote_http_api/server.clj
```diff
@@ -18,5 +18,6 @@
 
 (defn -main [& args]
   (let [rabbit-resources (queue/initialize)]
-    (.addShutdownHook (Runtime/getRuntime) (Thread. shutdown))
+    (.addShutdownHook (Runtime/getRuntime)
+                      (Thread. #(shutdown rabbit-resources)))
```
gws: :+1:

PR #14, opened 2018-02-28
Make ping responses text/plain

cap10morgan:
Otherwise they default to application/edn and show up as base64-encoded binary in some dev tools, browsers try to download them instead of just displaying them, etc.

This fixes that.


src/turbovote_http_api/service.clj
```diff
@@ -22,7 +22,8 @@
 
 (def ping
   (pedestal/interceptor
-   {:enter
+   {:name ::ping
```
theJohnnyBrown: Is adding the name here relevant to this content-type change?

theJohnnyBrown:
I know it's super small but this is the kind of thing that's easy to test so we might as well add one

theJohnnyBrown:
Overall LGTM 👍

theJohnnyBrown:
👏 👏 yay tests!

PR #15, opened 2018-07-09
Add endpoint for getting all states' voter reg. methods

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/153681289)


cap10morgan:
Depends on these PRs too:

- https://github.com/democracyworks/election-authority-works/pull/54
- https://github.com/democracyworks/voter-registration-works/pull/67

test/turbovote_http_api/elections_service_test.clj
```diff
@@ -7,25 +7,13 @@
...
             [turbovote-http-api.test-utils :as utils]
-            [turbovote-http-api.authorizer.dummy :as dummy]
             [turbovote-http-api.handlers.elections :as h.elections]
             [clojure.string :as str]
             [ring.util.codec :refer [form-encode]])
   (:import (java.util UUID)))
 
 (def root-url "/elections")
 
-(def use-dummy-authorizer
```
theJohnnyBrown: Any reason not to use [with-redefs](https://clojuredocs.org/clojure.core/with-redefs) for testing instead of adding a dynamic variable to the code?

src/turbovote_http_api/handlers/utils.clj
```diff
@@ -4,13 +4,20 @@
...
    [io.pedestal.interceptor :refer [interceptor]]
+   [turbovote-http-api.authorizer.three-scale :as three-scale]
+   [turbovote-http-api.boundary.authorization :as authorization]
    [turbovote.resource-config :refer [config]])
   (:import (java.time ZonedDateTime ZoneId)
            (java.time.format DateTimeFormatter)
            (java.util UUID)
            (org.apache.commons.codec.digest DigestUtils)))
 
+(def ^:dynamic *authorizer*
```
theJohnnyBrown: Does this need to be dynamic?

src/turbovote_http_api/handlers/voter_registration.clj
```diff
@@ -39,6 +39,17 @@
...
                     (utils/add-cache-headers)
                     (utils/promote-to-body :registration-methods)]
+    {:get [::get-state-registration-methods (bifrost/interceptor
+                                             (utils/ttl-memoize-async
+                                              read-registration-methods)
+                                             (utils/timeout-config
+                                              [:voter-registration
+                                               :read-registration-methods]))]}]
+   ["/registration-methods"
+    ^:interceptors [utils/authorize-api-request
```
theJohnnyBrown: Why is this endpoint behind authorization? (Actually I'm all little unclear on the motivation for this endpoint overall)

test/turbovote_http_api/elections_service_test.clj
cap10morgan: No one here likes `with-redefs`. :)

In fact, I think it originally _did_ use `with-redefs` and this refactor was suggested as an improvement to that.

src/turbovote_http_api/handlers/voter_registration.clj
cap10morgan: That's explained in the linked Pivotal card, but probably lacks sufficient context.

When we open up a new API endpoint to third-party access, we want to make sure that only valid API keys can access those endpoints but also that we don't break access for turbovote-web clients running in browsers (which do not have API keys). Typically the way this plays out is that the turbovote-web endpoint is more specific (e.g. give me this resource for this user / ocd-id / state) whereas the third-party endpoint is more general (e.g. give me this resource for all states). So we can get away with putting the authn|z in front of a new, more general endpoint and leaving it off of the more specific one for turbovote-web. It's not the _best_ solution, but it was deemed sufficient for this initial period of rolling out API access to a handful of large, well-known partners with whom we have good communication channels.

src/turbovote_http_api/handlers/utils.clj
cap10morgan: I think so? This code isn't actually being changed by this PR, just moved to a shared namespace so that more than one endpoint can use it.

test/turbovote_http_api/elections_service_test.clj
theJohnnyBrown: Well when you put it like that, point taken. It's not a big difference.

PR #22, opened 2018-09-14
Handle sms responses

eggsyntax:
https://www.pivotaltracker.com/story/show/159638701

Since we're switching to sending sms from shortcodes, we need to process stop and start sms messages within TV. 

Evaluate the body of each sms we receive and determine whether it qualifies as a stop or start message. Pass it accordingly to `election-notification-works` services `delete-by-phone` or `restart-by-phone`. If it's neither, send it on to zendesk. 

We expose this as an endpoint in our API, and twilio will call it on each sms it receives at one of our numbers. longcode messages exactly matching certain keywords will still be processed and replied to by twilio, but we un-/re-subscribe them anyhow. We return a message (formatted as twilio's TwiML) which twilio will send as a reply to the original sms.


src/turbovote_http_api/handlers/sms.clj
```diff
@@ -6,17 +6,88 @@
...
+  (el-notys/delete-subscription-by-phone {:phone (:From sms)}))
+
+(defn restart-subscription-by-phone [sms]
+  (el-notys/restart-subscription-by-phone {:phone (:From sms)}))
+
+(def message-actions {:stop delete-subscription-by-phone
+                      :start restart-subscription-by-phone})
+
+(def preferred-replies {:stop "You have successfully unsubscribed. You will no longer receive SMS notifications from TurboVote. Reply START to resubscribe."
+                        :start "You have successfully re-subscribed. Reply STOP to unsubscribe again."})
```
theJohnnyBrown: How feasible would it be to i18n these?

src/turbovote_http_api/handlers/sms.clj
```diff
@@ -6,17 +6,88 @@
...
+  the details of their message."
+  [sms result]
+  (let [message (:Body sms)
+        message-type (sms-utils/message-type message)]
+    (if (ok? result)
+      (if (skip-response? sms)
+       (text-to-twiml "") ; tells twilio we don't want to respond ourselves
+       (text-to-twiml (preferred-replies message-type)))
+     (do
+       (log/warn "We were unable to process a start/stop because we encountered:\n" result "\non sms:\n" sms)
```
theJohnnyBrown: I feel like log/error would be appropriate here

src/turbovote_http_api/sms_utils.clj
```diff
@@ -0,0 +1,88 @@
...
+
+(defn standard-stop-message?
+  "Does message contain one of the standard stop words?"
+  [s]
+  (some #(re-find (re-pattern %) s) standard-stop-words))
+
+(defn extended-stop-message?
+  "Does message exactly match one of our extended stop words?"
+  [s]
+  (some #(re-matches (re-pattern %) s) extended-stop-words))
```
theJohnnyBrown: For this function and `standard-stop-message?`, I feel like a better approach would be to tokenize the message (e.g. split on `\s+`) and then check the intersection with the relevant word set.

src/turbovote_http_api/handlers/sms.clj
eggsyntax: Not sure -- I haven't dealt with i18n in TV at all. I propose adding it as a TODO for now and addressing it once it's in prod, due to the time crunch.

src/turbovote_http_api/sms_utils.clj
eggsyntax: I considered that, but @rackette and I had some extended discussion on it, and ended up here -- we're trying to catch the common errors that they've seen end up in zendesk, and that includes some spacing weirdnesses (eg `"s t o p"`) and multiword statements (eg `"ALTO – TurboVote"`).

src/turbovote_http_api/sms_utils.clj
```diff
@@ -0,0 +1,88 @@
...
+  (with-standard-text-munging is-start-message? sms-body))
+
+(defn twilio-recognizable-message?
+  "Does message exactly match one of Twilio's standard stop or start words? This
+  is useful to know because Twilio will automatically respond to such messages."
+  [s]
+  (letfn [(twilio-match? [txt]
+            (some #(re-matches (re-pattern %) txt)
+                  (clojure.set/union standard-stop-words standard-start-words)))]
+    (with-standard-text-munging twilio-match? s)))
```
mrichards42: since these are exact matches (and you've already trimmed and lower-cased the message body) this could be just a check for set membership

src/turbovote_http_api/sms_utils.clj
```diff
@@ -0,0 +1,88 @@
...
+            [clojure.tools.logging :as log]))
+
+;;; Stop messages
+
+;; per https://support.twilio.com/hc/en-us/articles/223182208-Industry-standards-for-U-S-short-code-HELP-and-STOP
+;; Note: "END" is one of Twilio's standard ones, but we put it in the
+;;   extended-stop-words category (ie we treat only an exact match as a stop
+;;   request) because it's especially likely to cause false positives, eg
+;;   "Thanks for sending this!"
+(def standard-stop-words #{"stop" "cancel" "unsubscribe" "quit"})
```
mrichards42: I did a little bit of analysis on the all of the actual sms responses we've gotten and I have a list of false positives that this creates (by no means an exhaustive list). I think it would be worth a conversation with @rackette to see which of these false positives are acceptable. Some of these are probably easy to deal with (while keeping a partial match algorithm), while other are a bit harder.

I'll send the list along tomorrow, but here are a few categories:
* AutoReply messages from people who are driving (but who will reply once they've **stop**ped)
* Apparently some phones do "reactons" to texts? So we get messages like "Liked -- Thank you for being a voter! ... To **unsubscribe** reply **STOP** - TurboVote"
* Names or places: "Chri**stop**her", "mez**quit**e" (maybe this should be a word-boundary search instead of a substring match?)
* Legit zendesk questions, along the lines of "I requested an absentee ballot, but I'm actually going to vote in person. Can I cancel the absentee ballot request?"

eggsyntax: Yeah, this definitely results in potential false positives (although I didn't know about the reaction thing, jeez). @theJohnnyBrown makes a similar point [above](https://github.com/democracyworks/turbovote-http-api/pull/22#discussion_r218233185). My impression was that that was considered an acceptable risk; I'll tag @rackette today to talk about it more -- do you want to be included in that discussion? Until I'm more knowledgeable about the project, I'm totally happy to leave it to her and/or you to choose the appropriate balance between false positives, false negatives, and programmer effort.

src/turbovote_http_api/sms_utils.clj
eggsyntax: @theJohnnyBrown just as an FYI follow-up, talked more w/ Esha and Mike, & we decided to scale back to minimize false positives; we're now looking only for exact matches. Can look at more sophisticated approaches after NVRD.

src/turbovote_http_api/handlers/sms.clj
```diff
@@ -6,17 +6,90 @@
...
+  [sms]
+  (and (sms-utils/twilio-recognizable-message? (:Body sms))
+       (from-longcode? sms)))
+
+(defn- text-to-twiml
+  "Create an http response containing a message encoded in twilio's TwiML
+  format. An empty or nil argument will cause a message telling twilio to do
+  nothing."
+  [body]
+  (let [twiml (if (seq body)
```
mrichards42: I think this would be clearer if it were `(if-not (string/blank? body)`

src/turbovote_http_api/server.clj
```diff
@@ -13,11 +13,13 @@
...
+             {::http/join? false}) ; don't hang
       http/create-server))
 
 (defn -main [& args]
   (let [rabbit-resources (queue/initialize)]
     (.addShutdownHook (Runtime/getRuntime)
                       (Thread. #(shutdown rabbit-resources)))
-    (http/start (http-server))))
+    (http/start (http-server))
+    (prn "Started http server."))) ; Would otherwise return the (large) service map
```
mrichards42: Is returning the service map problematic? I could see it being helpful to have the service map returned if you're running `-main` in the repl (to run `http/stop` in particular).

src/turbovote_http_api/sms_utils.clj
eggsyntax: Decision (for posterity): scale back to exact matches for now; consider more sophisticated solutions post-NVRD.

src/turbovote_http_api/server.clj
eggsyntax: Only in that you get a whoppin-big data structure printed at the repl. I could go either way.

mrichards42: Eh, if it's a matter if one preference over another I'd rather leave it how it was, and if you don't like the huge print you could `(def server (-main))` instead.

Then again maybe it's just that I've gotten used to all kinds of junk being logged at my repl since fireplace has some trouble with broken cider sockets and dumps useless errors all the time 😬

src/turbovote_http_api/handlers/sms.clj
eggsyntax: :man_shrugging: [Documentation](https://clojuredocs.org/clojure.core/seq) and [style guides](https://github.com/bbatsov/clojure-style-guide#nil-punning) consistently say that `(seq x)` is the best idiom for checking whether you have a non-empty sequence, and a string is ultimately a sequence in Clojure.

But I can see your argument too, since it's a string (or nil) in particular; happy to switch it.

mrichards42: 👍 totally agreed if it were a non-string seq. I know you _aren't_ using it this way in code, but sombody _could_ do this:
```clj
(twilio-response "     ")
```
and then twilio might respond with blank text message

PR #24, opened 2018-09-25
Handle sms responses improvements

eggsyntax:
- make zendesk fire-and-forget
- better logging
- more tests
- refactoring to cleaner implementation


src/turbovote_http_api/handlers/sms.clj
```diff
@@ -17,7 +17,17 @@
...
+
+(defn zendesk
+  "Fire a message off to zendesk. For consistency with the other functions
+  called for sms messages (which are async->fn), returns a channel which will
+  receive an :ok result."
+  [message]
+  (let [response-chan (async/chan 1)]
+    (zendesk-fire-and-forget message)
+    (async/>!! response-chan {:status :ok})
+    response-chan))
```
eggsyntax: Note -- slightly weird pattern to return a channel that just gets :ok after a fire-and-forget. I'm totally open to feedback on that. The advantage is that it keeps the `sms-router*` code much cleaner, because it gets the same sort of response from every server.

The kehaar docs are slightly inconsistent about whether it's reasonable to use the output from the fire-and-forget function, so I erred on the safe side of ignoring it.

src/turbovote_http_api/handlers/sms.clj
```diff
@@ -60,36 +70,50 @@
...
+     (assoc ctx :response
+            (cond
+              (ok? result) (do (log/debug "Processed an sms of type"
+                                          (name message-type) ":" (pr-str body))
+                               (if (skip-response? sms)
+                                 no-response ; tells twilio we don't want to respond ourselves
+                                 (text-to-twiml (preferred-replies message-type))))
+              (nil? result) (do (log/debug "When handling sms, no response from service"
+                                           message-type "sms was" (pr-str sms))
+                                no-response)
```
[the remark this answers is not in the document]
eggsyntax: That seems sensible. I did the refactoring in response to various comments, but I can't imagine why we _wouldn't_ want to send to zendesk if our attempt to stop/start failed to respond. Changing.

src/turbovote_http_api/handlers/sms.clj
eggsyntax: Hmm, can't seem to reply directly to your comment ("I agree that it's a little weird..."); hopefully this'll look reply-ish enough to be clear.

The complication there is that "send to zendesk" is often a part of the standard flow rather than a fallback for errors; having it return the fake response is more about making it work cleanly as part of that standard flow. I suppose I could make both an `async->fn` version and a `async->fire-and-forget-fn`, but that doesn't feel very clean either.

mrichards42: Yeah, GitHub threading isn't the best. I was intending that as an overall comment, but we can keep the discussion here if that makes more sense.

I'm suggesting the standard flow might be "recognize a response and act on it", and the fallback is "send to zendesk if we either don't recognize the message, or encountered an error while processing". I'm wondering if that approach might make the code simpler.

mrichards42: Agreed -- if we got an error from election-mail-works we definitely want to log that loudly since that's definitely a different case than sending to zendesk b/c we didn't recognize the message.

I almost wonder if we should include a note in the zendesk message that indicates that we should have handled this, but something went wrong? Or maybe we _shouldn't_ send those messages to zendesk, but instead alert devs when a response isn't handled? I wonder if having a bunch of "STOP" messages in zendesk might just clog up the works?

eggsyntax: It might, but seems like we gotta get them unsubscribed, and if the code can't do it, it's probably gotta be a person...
I'll bump those cases to log/warn, though, so that they'll hopefully stand out more in the logs. Or we could certainly add a notification to devs, but I do think we should probably also send to zendesk.

See refactored version committed this morning.

mrichards42: I'm not sure my suggestion was clear. I don't mind at all if you disagree with the suggestion, but I'd like to make sure we're talking about the same idea.

I was aiming to get rid of the function that calls zendesk and then returns a channel with `{:status :ok}` since I think it makes the logic harder to follow -- in particular I think it makes the code read like if zendesk fails to process the message the first time, it will send the same message to zendesk again in the fallback. I realize it _won't_ do that b/c it's defined here to always return `:ok`, but as we've been discussing, that's unusual.

If `zendesk` isn't one of the `message-actions`, you wouldn't need the dummy-channel return, and it could look something like this:
```clj
(let [...
      result (when send-fn (async/<!! send-fn sms))]
  ;; ... log whatever needs to be logged ...
  ;; or maybe do the logging in send-fn?
  (if (ok? result)
    (assoc ctx :response ...)
    (do
      ;; Either no send-fn, or an error in the send-fn routes this to zendesk (f-a-f)
      (zendesk sms)
      (assoc ctx :response no-response))))
```

eggsyntax: Or to be a bit clearer, get rid of this pattern:
```
(let [send-fn (message-action-map message-type)
      result (async/<!! (send-fn sms))]
  ...)
```
where we assume that the action fn returns a channel that we need to take from.

mrichards42: I'm definitely open to more ways to direct traffic, especially since you've been continuing to think about this while working on the zendesk filtering card.

Another approach I thought of is using two pedestal interceptors: one for processing start and stop messages, and another for filtering and sending to zendesk (for any messages that weren't handled by the first interceptor). Pedestal might work well for this sort of "try all of these handlers until one succeeds" approach.

PR #25, opened 2018-10-17
Ignore some sms responses

eggsyntax:
https://www.pivotaltracker.com/n/projects/930636/stories/160565475/comments/195673500

[Note: PR shows changes from `handle-sms-responses-improvements` since that's not yet merged. Refactoring based on discussion in that PR has continued here.]

Now ignores certain types of SMS responses: emoji-only, simple thanks-only, and identifiable auto-replies.


mrichards42:
We might want to get some input from @tie-rack, since I know he was planning on looking into spinning `zendesk-works` off into a standalone service, and possibly using it as the fallback twilio response endpoint (for when the primary endpoint takes too long to respond). That makes me wonder if filtering ignorable messages might be better served by `zendesk-works`, but I can also see an argument for keeping all the message-contents-based routing in one place.

src/turbovote_http_api/sms_utils.clj
```diff
@@ -85,11 +113,25 @@
...
+  "Does string exactly match one of the patterns that don't need a response?"
+  [s]
+  (some #(re-matches (re-pattern (clojure.string/lower-case %)) s)
+        ignorable-patterns))
+
+(defn ignorable-message?
+  "Is the trimmed/lowercased message an ignorable message?"
+  ;; Note that emojis are ignored separately, in with-standard-text-munging.
+  [sms-body]
+  (with-standard-text-munging is-ignorable-message? sms-body))
```
mrichards42: Now that there are more categories, this pattern might be worth refactoring:
```clj
(defn is-x-message?
  [s]
  (some #(re-matches (re-pattern (clojure.string/lower-case %)) s)
        x-patterns))

(defn x-message?
  [sms-body]
  (with-standard-text-munging is-x-message? sms-body))
```

The first thing that comes to my mind is something like this:
```clj
(defn message-type
  [sms-body]
  (let [munged (munge sms-body)]
    (cond
      (matches-any? start-patterns munged) :start
      (matches-any? stop-patterns munged)  :stop)))
```

src/turbovote_http_api/sms_utils.clj
```diff
@@ -3,6 +3,34 @@
...
 
+;;; Text munging (applied consistently across all fns)
+
+(def really-emo-regex
+  "Recognizing emojis turns out to be a bit tricky, for various reasons (java
+  unicode handling, surrogate pairs, etc). So we strip them not by trying to
+  exactly recognize them, but by whitelisting the sorts of chars we want to
+  keep: unicode types L (letters from any language), N (numbers), P
+  (punctuation), and Z (whitespace)."
+  #"[^\p{L}\p{N}\p{P}\p{Z}]")
```
mrichards42: I like this aggressive approach to stripping out junk from messages. Should we be more aggressive and also strip out punctuation?

src/turbovote_http_api/sms_utils.clj
```diff
@@ -16,14 +44,9 @@
...
-(defn- with-standard-text-munging
-  "Apply f to trimmed & lower-cased version of sms-body, and return boolean"
-  [f sms-body]
-  (-> sms-body
-      string/trim
-      string/lower-case
-      f
-      boolean))
+(def ignorable-patterns #{#"^\s*$" ; whitespace-only (after stripping emojis & trimming)
+                          "Thanks", "Thanks!", "Thanks.", "Thx", "Thx!", "Thx.", "Thank you.", "Thank you!", "Thank you"
```
mrichards42: Turning those "Thanks" strings into patterns might make this more concise:
```clj
#"Thanks[.!]?"
```

src/turbovote_http_api/sms_utils.clj
```diff
@@ -16,14 +44,9 @@
                            "Syop" "Stpp" "atop"  "sto0" "st0o" "sopt" "dtip"
                            "remove" "alto"})
```
mrichards42: Two thoughts come to mind, neither of which I'm attached to:
1. The extended-stop and ignorable words are probably lists that we want to add to with some frequency. Does it make sense to store these (and the start/stop words, for consistency's sake) as config instead of code?
2. Is it worth lower-casing and compiling these into patterns here instead of on the fly when we test incoming sms bodies? idk if java caches compiled patterns (either way, probably a micro optimization, so not a big deal)

test/turbovote_http_api/handlers/sms_test.clj
```diff
@@ -0,0 +1,80 @@
+(ns turbovote-http-api.handlers.sms-test
+  (:require [turbovote-http-api.handlers.sms :as sut]
+            [clojure.core.async :as async]
+            [clojure.test :as t :refer [deftest is testing]]))
+
+(def sms-base {:AccountSid "testsid"
+               :From "+15555555555"
+               :To "88683" ; our shortcode
```
mrichards42: Anyone who gets texts from us will know what the short code is, so it isn't exactly private knowledge, but it might be better to use a dummy short code here. 🤷‍♂️

src/turbovote_http_api/sms_utils.clj
```diff
@@ -16,14 +44,9 @@
...
 
-(defn- with-standard-text-munging
-  "Apply f to trimmed & lower-cased version of sms-body, and return boolean"
-  [f sms-body]
-  (-> sms-body
-      string/trim
-      string/lower-case
-      f
-      boolean))
+(def ignorable-patterns #{#"^\s*$" ; whitespace-only (after stripping emojis & trimming)
```
mrichards42: Shouldn't `string/trim?` in `standard-text-munging` take care of whitespace-only texts?

src/turbovote_http_api/handlers/sms.clj
```diff
@@ -60,36 +67,59 @@
...
+           (nil? result)
+             (log/warn "When handling sms, no response from service"
+                        message-type "; sms was" (pr-str sms))
+           (= :error (:status result))
+             (log/warn "We were unable to process sms " (pr-str sms)
+                       " because we encountered: " (pr-str result))
+           ;; something strange is going on
+           :else
+             (log/warn "Not sure what to do about sms" (pr-str sms)
+                       "because:" (pr-str result)))
```
mrichards42: I'm not set on this, but since you're doing the `async/<!!` call in each of the `e-n-w` handlers, what about putting all this logging in those handlers as well? It might make the logic of this function clearer if the logging is elsewhere.

I've seen functions like this in a couple repos:
```clj
(defn fetch-some-async-resource [kw ch]
  (let [result (<!! ch)]
    (cond
      (= :ok (:status result) (kw result)
      (nil? result) (log/debug "Error fetching" kw ": timeout")
      :else (log/debug "Error fetching" kw ":" (pr-str result)))
```
Not that this is a perfect match, but it's close.

src/turbovote_http_api/handlers/sms.clj
```diff
@@ -17,25 +17,32 @@
 
 (defonce zendesk-channel (async/chan 1000))
 
-(def zendesk (wire-up/async->fn zendesk-channel))
+(def zendesk-fire-and-forget (wire-up/async->fire-and-forget-fn zendesk-channel))
```
mrichards42: 👍 Glad you were able to refactor this to get rid of the dummy channel from the other PR

src/turbovote_http_api/handlers/sms.clj
```diff
@@ -17,25 +17,32 @@
...
 
 (defn restart-subscription-by-phone [sms]
-  (election-notifications/restart-subscription-by-phone {:phone (:From sms)}))
+  (async/<!! (election-notifications/restart-subscription-by-phone {:phone (:From sms)})))
+
+(defn ignore
+  "Some messages we can simply ignore (eg auto-replies); they require neither
+  responses nor human attention."
+  [sms]
+  {:status :ok})
```
mrichards42: This is better than the dummy channel return from zendesk, but it's still a little unexpected. What about something like this?
```clj
(when-not (= :ignorable message-type))
  (zendesk-fire-and-forget sms))
```

src/turbovote_http_api/sms_utils.clj
```diff
@@ -85,11 +113,25 @@
     (clojure.set/union standard-stop-words standard-start-words)
```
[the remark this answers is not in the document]
eggsyntax: Excellent point.

src/turbovote_http_api/sms_utils.clj
eggsyntax: 1. Yeah, I think that's a great idea. That raises the possibility that we probably *shouldn't* regexify the "Thanks" messages, that we should use plaintext as much as possible to make it easy for non-devs to edit the list.
2. I'm gonna call that one a premature optimization ;)

src/turbovote_http_api/sms_utils.clj
eggsyntax: Just covering edge cases -- the only one I recall offhand is actually the empty string, which you can get after you strip out emojis from an emoji-only text. IIRC I thought about specifically ignoring that, but figured it couldn't hurt to cover any sequence that resulted in whitespace-only. It's a tossup, though; I'll switch it if you think it'd be clearer.

src/turbovote_http_api/sms_utils.clj
mrichards42: Agreed that this should all be a post-election conversation

1. This conversation got me to re-read some of the code, and I came across a subtle bug:
> That raises the possibility that we maybe shouldn't regexify the "Thanks" messages, that we should use plaintext as much as possible to make it easy for non-devs to edit the list.

All of these strings are being turned into `re-pattern`s, and if we're going to do that we need to escape the metachars, otherwise the `.` in `"Thanks."` is going to match anything. I think I'd argue that either everything should be a regex to begin with, so we don't have to worry about escaping things (which raises the barrier to editing the list), or strings should be treated differently than patterns.

2. Probably yeah. I was also thinking that if we end up moving the patterns elsewhere, we'll need a `read-patterns` function, which could also do that, e.g.
```clj
(defn read-patterns [filename]
  (-> (slurp filename)
      string/split-lines
      (map (comp re-pattern string/lower-case))
      set))
```

src/turbovote_http_api/sms_utils.clj
mrichards42: Nope, you're totally right. I'm not sure what I was thinking when I wrote that comment :)

src/turbovote_http_api/sms_utils.clj
mrichards42: Let's have a conversation about this (maybe with more people) after the election. If we end up needing partial matches, I think I'd push to make everything a partial match and have some patterns look like `#"^stop$"` so that matching code works the same way. I think having different logic based on the type of message we're matching is going to make the code harder to understand in the long run.

eggsyntax: As I understood it, the motivation for having different logic was that Esha felt like she had good data about what stop messages people send, but not about start messages. But I agree that unification is ultimately a good idea.

Can't comment on your suggestion to leave things here & move on, but WFM.

PR #27, opened 2018-11-02
Bugfix for sending to zendesk

eggsyntax:
This is a one-line bugfix for the previous PR . That PR was reverted, so you're seeing all [the changes made in that PR](https://github.com/democracyworks/turbovote-http-api/pull/25) as well as the current one-liner. The only new change is the switch from `:response true` to `:response false` in config.edn, for a function that was refactored from a `wire-up/async->fn` to a `wire-up/async->fire-and-forget-fn`.

Note that this time around it's been tested more extensively, wiring up a phone number to a test twilio service and from there to a local copy of this branch, and sending sms messages to be sure they're handled properly.


eggsyntax:
Note -- added a 2nd one-liner which updates a test to match the updated behavior.

PR #30, opened 2018-11-16
Url lookup notys

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/159857928


test/turbovote_http_api/election_authorities_service_test.clj
```diff
@@ -0,0 +1,30 @@
...
+            [turbovote-http-api.handlers.election-authorities
+             :as h.elec-auth]
+            [clojure.string :as str]
+            [clojure.edn :as edn]
+            [io.pedestal.test :refer :all])
+  (:import (java.util UUID)))
+
+(def root-url "/election-authorities")
+
+(deftest read-preference-test
```
mrichards42: lookup-urls-test?

test/turbovote_http_api/election_authorities_service_test.clj
```diff
@@ -0,0 +1,30 @@
...
+  (testing "GET to /lookup-urls/:state returns 200 & lookup-urls on success"
+    (let [state "ma"
+          ocd "ocd-division/country:us/state:ma"
+          urls {:local-election-authority-lookup-url "elec.html"
+                :local-reg-authority-lookup-url "reg.html"}
+          authority-data (merge urls 
+                                {:other-keys "other-data"})
```
mrichards42: Unlike some of the other tests here, it looks like you're also testing that we don't return extra data, which is great! I think that could use either a comment, or a message in the `is` macro, like `(is (= urls read-data) "only includes lookup urls")`

PR #31, opened 2018-12-10
Auto-unsubscribe

eggsyntax:
Accept http calls from AWS SNS about bounced emails, and request deletion of the corresponding subscription from election-notification-works.

Note that I've got a temporary feature flag which prevents the code from actually deleting subscriptions. Once this code is deployed, we can add a subscription to SNS pointing to the now-exposed API, and watch logs for a day or two before flipping the feature flag.

Should be merged after [e-n-w PR 166](https://github.com/democracyworks/election-notification-works/pull/166), which creates & exposes the service that this PR calls.


src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,93 @@
...
+        bounce-type (get-in-json json-msg ["bounce" "bounceType"])
+        diagnostic-code (get-in-json json-msg ["bounce" "diagnosticCode"])
+        bounced-emails (mapv #(get % "emailAddress") recipients)]
+    (cond
+      (empty? bounced-emails) {:status :error, :message "Call from SNS contained no bounced emails"}
+      (not= bounce-type "Permanent") {:status :ok, :message "Ignoring transient email bounce"}
+      :else (let [debug-msg (str "Unsubscribing emails" bounced-emails
+                                 (when diagnostic-code (str "which bounced because" diagnostic-code)))]
+              (log/debug debug-msg)
+              (async/<!! (delete-subs-by-email-fn {:emails bounced-emails}))))))
```
mrichards42: I _think_ this is what [bifrost/fn-interceptor](https://github.com/democracyworks/bifrost/blob/master/src/bifrost/core.clj#L93) is designed for (a fn that returns an async channel).

It seems like it might be more in-line w/ the way some of the other endpoints are written if this were a chain of interceptors instead of a single interceptor that contains all the logic. Maybe an interceptor chain like:
```clj
[sns/parse-and-validate
 sns/handle-subscription-confirmation ;; short-circuits
 (bifrost/interceptor
  sns/unsubscribe-bounces
  (utils/timeout-config [:election-notification-works :email-bounce])]
```

test/turbovote_http_api/handlers/sns_test.clj
```diff
@@ -0,0 +1,91 @@
...
+                                        [:request :json-params :Message]
+                                        altered-message))
+
+;; There are more cases that could use tests, but we can't forge them effectively
+;; because we can't sign them ;P
+
+(deftest test-unsubscribe-aws-handler
+  ;; TODO this must be changed when we turn off DELETE-FOR-REAL
+  (let [happy-val {:status :ok,
+                   :message "Deleted all subscriptions for emails [\"egg@democracy.turks\"]"}]
```
mrichards42: Do we care about anything aside from the `:status` of the response? If not this could be a simpler test, and you wouldn't have to change it based on that feature flag

src/turbovote_http_api/service.clj
```diff
@@ -89,6 +107,37 @@
...
+               (prn ctx)
+               ctx)}))
+
+  (def error-causing-interceptor
+    "Inserts an error :response when entered, which causes the interceptor
+    chain to short-circuit and return immediately."
+    (interceptor
+     {:enter
+      (fn [ctx]
+        (assoc-in ctx [:response :status] 403))}))
```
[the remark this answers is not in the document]
eggsyntax: If you think they'll be useful, let's do that. If you have no objection, though, I'll handle that after these are finished rather than roping in a fourth repo on this ticket ;)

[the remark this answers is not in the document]
mrichards42: That's fine by me. I'd also be ok w/ those being in `dev-src` in this repo

test/turbovote_http_api/handlers/sns_test.clj
eggsyntax: The TODO is in error, actually; because this is the collated response, it won't change. But you're right, we probably don't care about anything but the status; I'll change it.

test/turbovote_http_api/test_utils.clj
```diff
@@ -1,18 +1,32 @@
 (ns turbovote-http-api.test-utils
-  (:require [clojure.core.async :as async :refer [<! >!]]
+  (:require [clojure.core.async :as async :refer [<! >! chan go close!]]
```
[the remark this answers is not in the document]
eggsyntax: From the looks of it, we were already referring some stuff, so I just added onto that. I think that's a nicer way to do it with core.async, or at minimum to refer it as `:a` -- when you use fns/macros from core.async, you almost always use a few of them near each other, so spelling out `async` adds a fair bit of visual noise.
That said, I agree it's more important to go with our common convention than to do it the way I prefer; I'll change it.

src/turbovote_http_api/service.clj
```diff
@@ -42,9 +42,25 @@
...
+  ;; Details at https://forums.aws.amazon.com/message.jspa?messageID=418160
+  (pedestal/interceptor
+   {:name ::set-content-type-json
+    :enter (fn [ctx]
+             (let [uri (-> ctx :request :uri)]
+               (if (clojure.string/ends-with? uri "election-notifications/unsubscribe-aws")
+                 (-> ctx
+                     (assoc-in [:request :content-type] "application/json")
+                     (assoc-in [:request :headers "content-type"] "application/json"))
+                 ctx)))}))
```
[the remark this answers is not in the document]
eggsyntax: Back to what I was saying about AWS being so sloppy that it's surprising that they won the cloud-platform wars ;P
First mover advantage, I guess.

src/turbovote_http_api/server.clj
```diff
@@ -11,14 +11,31 @@
...
+(defn -dev
+  "Start the system. Exactly like -main, except that a) we deref routes at
+  runtime, so changes to routes will be picked up whenever you evaluate
+  turbovote-http-api.service, and b) we don't block the calling thread."
+  ;; See http://pedestal.io/guides/developing-at-the-repl for details
+  [& args]
+  (init-rabbit)
+  (http/start (http-server
+               {::http/routes #(route/expand-routes (deref #'service/routes))
+                ::http/join? false})))
```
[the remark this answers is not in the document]
eggsyntax: Yeah, I was planning to bring it up to the team (along with announcing the new env vars), but it seemed like it made sense to start by creating a non-intrusive version first as a strawman.

test/turbovote_http_api/test_utils.clj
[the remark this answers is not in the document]
mrichards42: Cool, yeah I think it would be good to have a style guide for this kind of stuff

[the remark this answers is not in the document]
eggsyntax: FWIW, I'm usually pretty skeptical about mandatory style guides, and think they should be extremely minimal or (usually) nonexistent. The clojure-formatter thread I posted in Slack (it was actually this thread that made me think about it) points out some of the reasons why. The biggest reasons from my perspective are a) that it's extra mental context that the programmer has to maintain, and as Rich Hickey points out ("_nobody_ can juggle 100 balls") we don't have that much to spare. And b) they can feel oppressive; we're craftspeople, and craftspeople function best without a lot of rules :)

That said, I'd be 100% in favor of an automated standard with accompanying formatter; that way the mental-context issue just disappears. Formatters certainly don't cover everything (eg probably wouldn't cover this particular question), but they cover a lot of the basics.

Of course my preference is really subjective. But where preferences differ, I tend to err on the side of fewer rules and less structure.

2¢ ;)

project.clj
```diff
@@ -8,6 +8,7 @@
                  [org.clojure/tools.logging "0.4.1"]
                  [turbovote.resource-config "0.2.1"]
                  [ch.qos.logback/logback-classic "1.2.3"]
+                 [clj-http "3.9.1"]
```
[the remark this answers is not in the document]
eggsyntax: Not even in test, it looks like; I'll get rid of it. The earlier, more manual version of SNS auth needed it, and I missed deleting it.

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,93 @@
...
+  [msg]
+  (let [message-manager (SnsMessageManager. aws-region)]
+    (.parseMessage message-manager
+                   (java.io.ByteArrayInputStream. (.getBytes msg)))))
+
+(defn unsubscribe-aws-handler [delete-subs-by-email-fn ctx]
+  (let [request (:request ctx)
+        ;; TODO this line can be deleted once we turn the feature flag off
+        delete-subs-by-email-fn (if DELETE-FOR-REAL delete-subs-by-email-fn dummy-delete-fn)
+        json-params (-> request :json-params json/encode)
```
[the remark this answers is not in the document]
eggsyntax: Weirdly, the `:body` coming from SNS is a tagged literal:
`#object [org.eclipse.jetty.server.HttpInputOverHTTP 2100655193 "HttpInputOverHTTP@7d357459[c=0,q=0,[0]=null,s=STREAM]"]`

Have you encountered that before?

test/turbovote_http_api/test_utils.clj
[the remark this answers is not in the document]
mrichards42: I come down pretty heavily in favor of more guidelines = less to worry about (that's what linters are for) + less to argue about (e.g. one of the changes in this PR is replacing the 2-spaces-most-places indentation that vim uses with the 1-space-if-its-not-a-special-form indentation that emacs uses [here](https://github.com/democracyworks/turbovote-http-api/pull/31/files#diff-38cd75573d446fe4d2400be103b2824aL41))

I think consistency is important above all (and you definitely did a service here, since this file was using `<!` and `>!` but also `async/chan` and `async/go`), and if we care about being consistent in one place, why not have that established as a norm? I can see a pretty good argument for referring the async macros too (we use them so much that it isn't much extra overhead to think of them as part of the language)

I do think we're pretty well in agreement that guidelines shouldn't add more to think about. :)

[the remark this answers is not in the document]
eggsyntax: >  replacing the 2-spaces-most-places indentation

Heh, that one wasn't deliberate. Explicitly referencing async changed enough indentation that I just did indent-file.  :blush: 
I can switch that back.
& yeah, to the extent that it doesn't require human attention, I'm all in favor of unified style.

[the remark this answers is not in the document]
mrichards42: don't worry about that one -- li moved the function to utils in her PR :)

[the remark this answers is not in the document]
eggsyntax: LOL

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,93 @@
...
+  [json path]
+  (let [edn (json/parse-string json)]
+    (get-in edn path)))
+
+(defn unsubscribe-bounces
+  "Unsubscribe the bouncing emails described in an aws-response (excepting
+  transient bounces)."
+  ;; Note - can include complaints as well as bounces; we handle them the same
+  [delete-subs-by-email-fn sns-message]
+  (let [json-msg (.getMessage sns-message)
```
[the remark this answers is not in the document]
eggsyntax: 1. It's kind of obnoxious, isn't it? The whole hullabaloo just gets us back to the original :Message in the json-params/body-params. The difference is that now it's a fully authenticated version which we know hasn't been corrupted. I'll look for a way to clarify that.
2. Yeah, you mean the json parsing that happens in `get-in-json`? That's a good point, it's enormously more sensible to just convert the silly thing to edn.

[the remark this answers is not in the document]
mrichards42: 1. I don't think that's quite it, right? isn't the request body something like
```clj
"{\"Message\": \"{\\\"some-key\\\": \\\"some-value\\\"}\"}"
```
and then the `:json-params` (or whatever that is) are a decoded version of that:
```clj
{:Message "{\"some-key\": \"some-value\"}"}
```
but the `:Message` itself is still json, which is why you need to decode it again (or maybe that's what you were saying).

2. 👍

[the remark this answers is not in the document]
mrichards42: but right, fundamentally what we're getting out of `parse-message` is authentication, and the fact that the payload ends up wrapped in a class is a little obnoxious, but whatever, since using the library for authentication is so helpful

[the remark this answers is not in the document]
eggsyntax: Too many overloaded terms :laughing: 

The body (see also earlier discussion thread) is a tagged literal,
`#object [org.eclipse.jetty.server.HttpInputOverHTTP 2100655193 "HttpInputOverHTTP@7d357459[c=0,q=0,[0]=null,s=STREAM]"]`

Have you encountered such a beast before? I haven't. I know I did a bit of googling & it wasn't obvious that there was a useful way to deal with it, although I don't recall it. So by the time we see the context, it's all EDN. That said, it *does* contain a `:Message`, which is itself json-encoded. But we have to pass the message-manager a json-encoded version not of the :Message, but of the whole collection of params, so we have to re-encode them.

Does that make sense? Honestly, I only intermittently have the whole structure in my head; it's fairly baroque...

Another way to say it, which you may be better off ignoring unless you really want to get into the weeds, is that IF the message is uncorrupted,

```
(let [json-params (:json-params request)]
  (= (:Message json-params)
     (.getMessage
      (.parseMessage message-manager
                     (java.io.ByteArrayInputStream.
                      (.getBytes (json/encode json-params)))))))
```

src/turbovote_http_api/server.clj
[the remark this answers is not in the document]
eggsyntax: Yeah, makes sense, this one's already too big :laughing:

src/turbovote_http_api/handlers/sns.clj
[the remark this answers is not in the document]
mrichards42: oh, I didn't realize that the body was a stream, and it looks like it's [consumed](https://github.com/pedestal/pedestal/blob/67b5be563f80d220d2ed6c18775d84dc093e1763/service/src/io/pedestal/http/body_params.clj#L115) when json is parsed out of it. I suppose you could use a custom no-op body-params parser that left the stream alone and then let `SnsMessageManager` consume the stream (since as you found out, we are [strict](https://github.com/democracyworks/pedestal-toolbox/blob/c5b4505613fa73ef804725b7b23cc93b2a323a6b/src/pedestal_toolbox/params.clj#L45) about the content types we accept). But I don't have a problem with going through the decode/re-encode dance here.

All I was saying in the initial `1.` comment was that I expected that `:json-params` would already contain the deserialized sns message payload, but since that part is doubly-encoded, you have to parse it anyways.
As in, the request body looks like this:
```clj
(json/encode {:Message (json/encode {:whatever :data :is :here})})
```

src/turbovote_http_api/handlers/sns.clj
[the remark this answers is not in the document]
mrichards42: yup, you're right, as noted in the other thread

src/turbovote_http_api/handlers/sns.clj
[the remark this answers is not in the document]
eggsyntax: Yeah, totally.

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,110 @@
...
+        diagnostic-code (get-in edn-sns-message ["bounce" "diagnosticCode"])
+        bounced-emails (mapv #(get % "emailAddress") recipients)]
+    (cond
+      (empty? bounced-emails) {:status :error, :message "Call from SNS contained no bounced emails"}
+      (not= bounce-type "Permanent") {:status :ok, :message "Ignoring transient email bounce"}
+      :else ; proceed to unsubscribe each email
+      (let [debug-msg (apply str "Unsubscribing emails " bounced-emails
+                             (when diagnostic-code (str " which bounced because " diagnostic-code)))
+            delete-responses (for [email bounced-emails]
+                               (async/<!! (delete-subs-by-email-fn {:email email})))
```
mrichards42: I hate to say this now that you've rewritten this entire function, but given how many lines you had to add here, and that they're all very similar to the code you have in [e-n-w](https://github.com/democracyworks/election-notification-works/pull/166/files#diff-fbeb90745da5b660bd9190000a691b61R97) . . . I wonder if it's worth having the e-n-w queues take both singular and plural forms? I still don't think it warrants 2 separate queues.

like
```clj
(defn delete-users-by-email*
  [users-by-email-fn delete-subscription-fn {:keys [email emails] :as message}]
  (s/validate by-email-schema message)   ;; obviously the schema would have to change
  (let [emails (or emails [email])       ;; but the only changes in the function are this ...
        users (mapcat users-by-email-fn emails)] ;; ... and adding mapcat back
    ...))
```

mrichards42: Summarizing a convo in slack about this: most of the added complexity here is for logging purposes, and the backend component also logs all the requests and results, so another possibility for simplifying this code would be to remove all the extra logging.

src/turbovote_http_api/handlers/sns.clj
eggsyntax: I've spent some time thinking about this & trying stuff out, and my feeling is that pulling the parse-and-validate step out as a separate interceptor is a great idea, but the rest of it is better left as-is. This thing is a PITA to understand, and it's naturally modeled as a branching decision tree, and writing that as a series of short-circuiting interceptors makes that less clear. That's especially true because a) the munging and exception handling are both finicky, and b) handling complaints adds another step to the decision tree. I'm all for following conventions, but I think this is a case where the most important thing is to model this complicated thing in the most natural way, for the sake of the next person who has to deal with it.

mrichards42: 👍 for sure on at least pulling `parse-and-validate` (or whatever you'd like to call that) into its own interceptor.

What I'm getting at is that there's a lot of aws/sns boilerplate, and it would be nice if we didn't have to repeat all this boilerplate if we ever want to handle sns messages again. And you're right, this is a branching decision tree (and the java classes model that to a certain extent), so while you could model it as an interceptor chain, it might not be the best fit.

What if the `sns` namespace were responsible for handling all the tough parts of sns, and the business logic of handling what to _do_ with the message went elsewhere? So like . . .
```clj
;; all in the election-notifications namespace

;; the business logic (which you already have)
(defn emails-to-unsubscribe
  [{:keys [bounce complaint] :as already-parsed-notification-no-json-here}]
  ...)  

(defn unsubscribe-bounces-or-complaints
  [unsubscribe-fn sns-noty]
  (let [emails (emails-to-unsubscribe sns-noty)]
    ...))

;; I don't even have to think about sns here
(sns/notification-interceptor
  (partial unsubscribe-bounces delete-subscriptions-by-email))
```

eggsyntax: That makes sense. I've taken some steps in that direction now (see what you think), and am inclined to leave it there for now until/unless we see an additional use case. We could certainly go further with `sns-handler` (formerly `unsubscribe-aws-handler`) but it feels premature at this point.

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,127 @@
+(ns turbovote-http-api.handlers.sns
+  "Handle http calls from AWS Simple Notification Service, typically representing
+  notifications of bounced emails. Documentation:
+  https://docs.aws.amazon.com/sns/latest/dg/sns-http-https-endpoint-as-subscriber.html
+  https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html"
```
mrichards42: I'm not sure what the right name for this is, but I don't think this namespace should be called `sns`. The code in this ns is pretty specific to using SNS to route SES bounces + complaints to us, but afaik you can use SNS for all kinds of things.

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,127 @@
...
+  "Unsubscribe the bouncing emails described in an aws-response (excepting
+  transient bounces)."
+  ;; Note - There seems to only ever be one bouncing email per SNS notification,
+  ;;        but we assume multiple to adhere to their standard.
+  [delete-subs-by-email-fn sns-message]
+  (let [inner-message (sns-utils/get-inner-message sns-message)
+        {:strs [bouncedRecipients bounceType diagnosticCode]} (get inner-message "bounce")
+        bounced-emails (mapv #(get % "emailAddress") bouncedRecipients)]
+    (cond
+      (empty? bounced-emails) {:status :error, :message "Call from SNS contained no bounced emails"}
```
mrichards42: Is this actually an error?

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,127 @@
...
+      (not= bounceType "Permanent") {:status :ok, :message "Ignoring transient email bounce"}
```
mrichards42: Similar to some of the other `{:status :ok, :message "..."}` responses that we talked about earlier -- if this message is important we should log it, but otherwise afaik nothing happens with this `:message`.

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,127 @@
...
+          ;; Minimal responses, since we don't do anything with them
+          {:status :ok}
+          {:status :error, :type :semantic}))
+      {:status :error, :message "Call from SNS contained no emails associated with complaints."})))
+
+(defn handle-bounce-or-complaint [delete-subs-by-email-fn sns-message]
+  (let [inner-message (json/parse-string (.getMessage sns-message))]
+    (case (get inner-message "notificationType")
+      "Bounce" (unsubscribe-bounces delete-subs-by-email-fn sns-message)
+      "Complaint" (unsubscribe-complaints delete-subs-by-email-fn sns-message))))
```
mrichards42: you could avoid another `(json/parse-string (.getMessage sns-message))` if you passed the `inner-message` into these functions instead of the `sns-message`

src/turbovote_http_api/handlers/sns.clj
```diff
@@ -0,0 +1,127 @@
...
+                   (condp = (type sns-message)
+                     SnsSubscriptionConfirmation (sns-utils/confirm-subscription! sns-message)
+                     SnsNotification (handle-bounce-or-complaint delete-subs-by-email-fn sns-message)
+                     (do (log/error "Unable to handle this request from AWS SNS:" json-params)
+                         {:status :error, :type :semantic}))
+                   (catch Exception e
+                     (log/error "Unexpected error while attempting unsubscribe on " sns-message ":" e)
+                     {:status :error,
+                      :type :server,
+                      :message "Unexpected internal error"}))]
```
mrichards42: Another `:message` that's thrown away I believe

src/turbovote_http_api/sns_utils.clj
```diff
@@ -0,0 +1,47 @@
...
+           message-manager (SnsMessageManager. aws-region)]
+       (try
+         (assoc ctx :sns-message
+                (.parseMessage message-manager
+                               (java.io.ByteArrayInputStream. (.getBytes json-msg))))
+         (catch Exception e
+           (log/error "Unparseable or inauthenticable message from SNS:"
+                      json-params "  ->  "
+                      (.getMessage e))
+           (assoc ctx :response {:status :error, :type :validation})))))})
```
mrichards42: I may be wrong on this, but I don't believe that this (or the responses in `sns-handler` for that matter) are ever converted to actual ring responses, so pedestal won't actually short-circuit this (since as Li found out the hard way, pedestal checks for a ring response with [this function](https://github.com/ring-clojure/ring/blob/900aaffb364dc97d25d91e97abd508b8dc3255cb/ring-core/src/ring/util/response.clj#L236)). As Li and I found out in her PR, there's a util function that creates a valid response [here](https://github.com/democracyworks/turbovote-http-api/blob/4843b2c455166b0eb7c6c05c12f267f4d866747f/src/turbovote_http_api/handlers/utils.clj#L43)

src/turbovote_http_api/handlers/election_notifications.clj
```diff
@@ -38,17 +43,21 @@
...
+        (if (not= (::utils/status unsigned-token) :error)
+          (-> (assoc-in ctx [:request :path-params :user-id] user-id)
+              (assoc-in  [:request :path-params :mediums] #{:email})
+              (update-in [:request :path-params] dissoc :user-token))
+          (assoc-in ctx [:response :status]
+                    403))))}))
+
+(def unsubscribe-aws-interceptor
+  (interceptor
+   {:enter (partial sns/sns-handler delete-subscriptions-by-email)}))
```
mrichards42: Similar to the comment about the `sns` namespace -- I think the name you had before (`unsubscribe-aws-handler`) was more accurate, since `sns-handler` sounds pretty general, while this handler is still pretty specific -- it's designed to unsubscribe bounced emails/complaints from SES that happen to come through SNS

src/turbovote_http_api/handlers/sns.clj
mrichards42: FWIW, this code is still pretty tied to handling SES bounces + complaints (i.e. I don't think it's any more general than it was before [this commit](https://github.com/democracyworks/turbovote-http-api/pull/31/commits/b0bb016b0904e0dad789d061367df74b21aa3f3c)), but I'm totally fine with that. The suggestions I was making in this thread were really just suggestions, and since this is the first time we've tried to do anything with SNS in this repo, (a) we don't know if we're going to need to do anything else with SNS, so any abstractions might be wasted effort, and (b) if we did try to make this super general (which yes, is what I've been advocating) we'd probably figure out later that we forgot to generalize function X, and that actually function Y that we did abstract really had no business being abstracted.

I think there's an argument to be made both ways, so if you're happy with where we've landed on this, I'm happy with it too

mrichards42: And just to be clear, I do think there's value in having the exclusively sns-related functions in their own namespace, so 👍 on that commit, if it looked like I was calling it out

src/turbovote_http_api/handlers/election_notifications.clj
eggsyntax: For sure. For a minute there I was thinking I was going to turn that into a generalized `sns-handler` that would delegate to various fns. Looks like I rushed that last commit a bit given the various comments above...

src/turbovote_http_api/handlers/sns.clj
eggsyntax: As I think about it again, I'm inclined to leave the messages in there (and add back the ones I took out). They don't do any harm, they could be helpful to someone reading the code, and they'll make our lives easier if we decide to turn on logging on the SNS side (which is a thing I saw somewhere that you can do). Thoughts?

mrichards42: Can you explain more about SNS logging?

eggsyntax: It's certainly not something I know much about, but SNS offers several kinds of logs. The one I think is most likely to be relevant is [delivery status logs](https://aws.amazon.com/premiumsupport/knowledge-center/troubleshoot-failed-sns-deliveries/), which "log your push notification transactions to Amazon CloudWatch Logs, logging both successful and failed deliveries," and include "the time taken (dwellTimeMs), message ID, destination endpoint, the response from the provider, and a status code." There are also logs for [SNS API calls](https://docs.aws.amazon.com/sns/latest/dg/sns-logging-using-cloudtrail.html) and [SNS topic monitoring logs](https://docs.aws.amazon.com/sns/latest/dg/sns-monitoring-using-cloudwatch.html), but at first look I don't think either of them is relevant (the latter may be, slightly, but it's coarse-grained monitoring).

I wouldn't propose turning it on at this point, since our internal logs should cover most failure modes, but if we run into tricky delivery problems I can imagine doing so, and at that point it seems like error messages from our handlers might be useful.

mrichards42: My objection is more that `:message` doesn't mean anything in this context, and I'd prefer that whatever information we want to convey here makes sense semantically, whether that's a comment, a log, or a response body.

src/turbovote_http_api/sns_utils.clj
mrichards42: What I mean here is that maps like `{:status :error :type :validation}` or `{:status :ok}` would work fine in the bifrost ecosystem, since bifrost takes care of converting those to ring responses. Since this code doesn't use bifrost, these functions should return the actual ring response itself.

E.g. this request should return 500, but instead it 404s:
```bash
$ curl -i -X POST localhost:8080/election-notifications/unsubscribe-aws -d '{}'
HTTP/1.1 404 Not Found
Date: Tue, 18 Dec 2018 17:12:26 GMT
Content-Type: text/plain
Transfer-Encoding: chunked

Not Found
```

eggsyntax: Oh, now I gotcha. Somehow I thought pedestal would convert to ring-responses. I'll dig into it, thanks.

src/turbovote_http_api/handlers/sns.clj
eggsyntax: OK, now that I understand what you were getting at, I've corrected them all to be normal pedestal/ring responses :)

src/turbovote_http_api/handlers/unsubscribe_aws.clj
```diff
@@ -0,0 +1,130 @@
...
+                                     :body "Ignoring transient email bounce"}
+      :else ; proceed to unsubscribe each email
+      (let [debug-msg (apply str "Unsubscribing emails " bounced-emails
+                             (when diagnosticCode (str " which bounced because " diagnosticCode)))
+            delete-responses (for [email bounced-emails]
+                               (async/<!! (delete-subs-by-email-fn {:email email})))]
+        (log/debug debug-msg)
+        (if (every? #(= (:status %) :ok) delete-responses)
+          ;; Minimal responses, since we don't do anything with them
+          {:status 200, :body "Unsubscribed bouncing emails."}
```
mrichards42: This needs `:headers {}`

Just a thought -- having to remember to add an empty `:headers` map to all the responses feels error-prone to me, and might lend itself to a function. We could factor the response creation out of [handlers.utils/assoc-response](https://github.com/democracyworks/turbovote-http-api/blob/4843b2c455166b0eb7c6c05c12f267f4d866747f/src/turbovote_http_api/handlers/utils.clj#L43) into something like
```clj
(defn response [status & [body]]
  {:status status
   :headers {}
   :body body})
```

src/turbovote_http_api/handlers/unsubscribe_aws.clj
```diff
@@ -0,0 +1,130 @@
...
+;; since this involves coordination with SNS functionality, I thought I'd play
+;; it safe.
+(def DELETE-FOR-REAL false)
+(def dummy-delete-fn (fn [& args]
+                       (let [ch (async/chan)]
+                         (async/go
+                           (log/debug "Only pretending to delete a subscription because feature flag is off.")
+                           (async/>! ch
+                                     {:status 200,
+                                      :body "Pretended to delete subscription"})
```
mrichards42: This one probably needs to be `{:status :ok}` 🙂

src/turbovote_http_api/handlers/unsubscribe_aws.clj
eggsyntax: Good idea on enriching `assoc-response`.

That said, I _think_ this doesn't need headers because it doesn't need to short-circuit.

mrichards42: Hmm, I think you're right (but then neither do any of the error responses in this file). Maybe instead of thinking of this as an interceptor we think of this as a [handler](http://pedestal.io/reference/handlers)?

src/turbovote_http_api/handlers/unsubscribe_aws.clj
eggsyntax: Dang, you're right :man_facepalming:

src/turbovote_http_api/handlers/unsubscribe_aws.clj
eggsyntax: It does seem like it would be natural to make this a handler, and I think it's clear how I'd change `unsubscribe-aws-handler`. What I haven't found yet is how that would change the route definition; if you happen to know that (or know where it's documented) it'd be useful info.

eggsyntax: Complicated slightly by the fact that the `parse-and-validate` interceptor assocs the extracted `sns-message` into the context. But I suppose it could assoc it into the request instead?

test/turbovote_http_api/handlers/unsubscribe_aws_test.clj
```diff
@@ -72,17 +72,17 @@
...
 (deftest test-unsubscribe-aws-handler
   (let [happy-val {:status :ok}]
-    (is (= :ok
+    (is (= 200
            (get-in (sut/unsubscribe-aws-handler (utils/simple-dummy-service happy-val)
                                                 ((sns-utils/parse-and-validate-message :enter) good-context))
                    [:response :status]))
         "Sadly, unsubscribe-aws-handler did not return the happy-val we wanted.")
-    (is (= {:status :error, :type :semantic}
+    (is (= {:status 400, :headers {}, :body "Invalid request, unknown SNS message type"}
```
mrichards42: these might be more resilient tests if they only checked the status, since the message could change :shrug:

mrichards42:
Heads up, it looks like your most recent [commit](https://github.com/democracyworks/turbovote-http-api/pull/31/commits/1972e7ee382553491e8a32c6b7fb27426ccb152e) introduced more changes than you meant to.

src/turbovote_http_api/handlers/unsubscribe_aws.clj
eggsyntax: Took a stab at this. Feels like a bit of a tossup. Intuitively it feels cleaner to me to assoc the parsed/validated sns-message into the ctx, not the request. But I'd love to get your thoughts; it's [here](https://github.com/democracyworks/turbovote-http-api/tree/unsubscribe-aws-throwaway).

eggsyntax: Hmm, failing test on that (passes in dev); I'll hold off on addressing that until I get your opinion on it.

mrichards42: Semantically I think it makes sense if this is a handler (since afaik there's no sense in putting another interceptor after it -- all paths return a response -- but it doesn't make much of a difference to me, and I'm fine if it's an interceptor.

I do think it make more sense for `parse-and-validate-message` to assoc its result on the `:request` since it's a transformation of the request (in the same way that `body-params` is a transformation of the request body). As a counter example, bifrost sticks an async channel on the ctx in `:enter`, then takes from that channel in `:leave` (so, using ctx for "private" data).

src/turbovote_http_api/service.clj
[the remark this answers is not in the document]
mrichards42: Whatever ends up happening w/ these, I'd prefer if they were either moved to `dev-src` or put in a different PR

src/turbovote_http_api/sns_utils.clj
```diff
@@ -0,0 +1,49 @@
...
+    SnsNotification or an SnsSubscriptionConfirmation. Will throw if message
+    can't be confirmed as an authentic, properly certified, signed call from AWS."
+     (let [json-params (-> ctx :request :json-params)
+           json-msg (json/encode json-params)
+           message-manager (SnsMessageManager. aws-region)]
+       (try
+         (assoc ctx :sns-message
+                (.parseMessage message-manager
+                               (java.io.ByteArrayInputStream. (.getBytes json-msg))))
+         (catch Exception e
```
mrichards42: Does SnsMessgeManager throw a particular exception when this happens?

src/turbovote_http_api/handlers/unsubscribe_aws.clj
mrichards42: I don't feel strongly enough about that to block this PR, but I would prefer if we waited to deploy this to production until the new year, since almost everyone is going to be out next week. But I'm 👍 on getting this deployed to the working/staging cluster, and then adding an SNS subscription to this so we get logs while we're away.

eggsyntax: I've been debating that myself. I was going to go for it since it only logs, but I trust your judgement on that :)

Do we have docs somewhere on deploying to and accessing that cluster, and ideally on where it exposes its endpoints? I don't know anything about except idle hearsay ;)

src/turbovote_http_api/service.clj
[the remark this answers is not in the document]
eggsyntax: Sure, I'll just pull them out for now.

PR #32, opened 2018-12-10
Returning user app state

li-at-dw:


test-resources/config.edn
```diff
@@ -0,0 +1,12 @@
+{:server {:hostname "0.0.0.0"
+          :port 58080
+          :allowed-origins [".*"]}
+ :three-scale {:service-token #resource-config/env "THREE_SCALE_SERVICE_TOKEN"
+               :service-id #resource-config/env "THREE_SCALE_SERVICE_ID"}
+ :twilio {:account-sid #resource-config/env "TWILIO_ACCOUNT_SID"}
+ :zendesk {:username #resource-config/env "ZENDESK_USERNAME"
+           :password #resource-config/env "ZENDESK_PASSWORD"}
+ :aws {:region "us-east-1"}
+ :jwt {:secret ""}
```
mrichards42: It isn't clear to me that a blank jwt secret will do any kind of encoding or validation (it probably does, but I don't know). It would be good to clear up any confusion by having that be a dummy, but not blank, value.

test/turbovote_http_api/election_notifications_service_test.clj
```diff
@@ -112,7 +114,22 @@
...
+                                            :mediums #{:email}}}
+                            #(= {:user-id fake-user-id
+                                 :mediums #{:email}} %))
+      (let [response (response-for
+                      utils/service :put
+                      (str/join "/" [root-url "unsubscribe" token]))
+            delete-data (edn/read-string (:body response))]
+        (is (= 200 (:status response)))
+        (is (= {:subscription {:user-id fake-user-id, :mediums #{:email}}}
+               delete-data))))))
```
mrichards42: another 👏  for adding tests that should have been there all along

src/turbovote_http_api/handlers/users.clj
```diff
@@ -47,4 +48,12 @@
...
-               (utils/timeout-config [:user :delete]))]}]])
+               (utils/timeout-config [:user :delete]))]}]
+   ["/returning-user/:user-token"
+    ^:interceptors [utils/validate-token-interceptor
+                    (bifrost.i/update-in-request [:path-params :user-id]
+                                                 [:path-params :id]
+                                                 identity)]
+    {:get [::get-returning-user (bifrost/interceptor
+                                 read
+                                 (utils/timeout-config [:returning-user :read]))]}]])
```
mrichards42: We might need to restrict what information is returned here. The jwt is _very_ unsecure (anyone who has the token will have access to whatever data we give them), but we're using it for convenience so that people don't have to create an account with a password.

I think the decision that was made around this problem was to give the user access a very limited amount of info, but allow them to _change_ whatever. So, e.g. they can update their address, but they can't look up the address we already have for them.

Now that I'm looking at this file though . . . we also have an unsecured get route on `/user/:user-id` so . . . that probably needs to go too

li-at-dw: Yup.  I have an open question out to Esha about what info should be sent to tv-web to be editable.  Apparently the old version of tv-web did allow users to look at their old address and edit the changes.

PR #36, opened 2019-01-14
Minor unsubscribe aws changes

eggsyntax:


src/turbovote_http_api/sns_utils.clj
```diff
@@ -18,7 +18,8 @@
   [confirmation]
   (log/debug "Confirming SNS subscription")
   ;; Make the GET request:
-  (.confirmSubscription confirmation))
+  (.confirmSubscription confirmation)
+  {:status 200, :body "Confirmed SNS subscription"})
```
li-at-dw: does it need the `:headers {}` bit here?  Just going by the way the other responses are written.

PR #37, opened 2019-02-27
Setup cloudamqp config for dev deployment

mrichards42:
* Upgraded kehaar so we can use the most recent standard queue intialization procedure (see [this commit](https://github.com/democracyworks/user-works/commit/7af1f6b719b52ed44945cba55d5cef54d84b8b2a#diff-37f56a9bcdddbd318f12c796a87c1f21) from user-works for what we should be standardizing on)
* Added additional rabbitmq env vars to resources/config.edn
* Added cloudamqp secrets and configmaps to both development and production deploys
  * These will be shared by all services that need access to cloudamqp in the dev ns
  * The production versions just copy the current config, in preparation for also moving prod to cloudamqp

Hopefully reactiveops can make sure the deploy stuff is set up properly, and gordon can double-check my work on the rest.


deploy/production/cloudamqp.configmap.yml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cloudamqp
+data:
+  RABBITMQ_PORT_5672_TCP_ADDR: "rabbitmq-ha.rabbitmq.svc.cluster.local"
```
gws: I'd consider calling these ConfigMap and Secret manifests `rabbitmq.{configmap,secret}.yml` since they contain RabbitMQ configuration and secrets -- they _can_ contain configuration that points services to CloudAMQP, but could also point services to a different cloud service or to an in-cluster RabbitMQ service (as is the case here).

gws: That would involve changing the names and references as well, of course.

mrichards42: that makes a lot of sense to me. I was going off of turbovote-admin-http-api, which uses `cloudamqp` as the name for these, but I'd be on board with changing these to be `rabbitmq`

deploy/development.config
```diff
@@ -22,6 +23,7 @@ POD_DISRUPTION_BUDGETS=(
 
 SERVICES=(
     development/turbovote-http-api
+    development/cloudamqp
```
mrichards42: oops, that should have gone under secrets, not services

PR #40, opened 2019-05-22
Remove code for no-longer-used turbovote-signup endpoint

eggsyntax:
Sister PR to https://github.com/democracyworks/election-notification-works/pull/182


eggsyntax:
Not doing an immediate deploy to prod since no behavioral consequences.

PR #43, opened 2019-07-02
Update to Java 11 and Clojure 1.10.1

gws:
In addition to updating these large bits of software, we can get some quick memory efficiency and configuration wins out of this as well.

**There is a Jetty security fix included with the Pedestal 0.5.7 updates, and so we should deploy and test this soon.**

The full list of improvements are:

* Java 11
* Clojure 1.10.1
* Removed New Relic APM agent
  * We are migrating off of this, so we might as well start removing it.
* No longer running as root by default
* Simplify lein usage in the Dockerfile
* No longer require `DEPLOYMENT_NAME`
* No longer require `CGROUP_HEAP`
* No longer require `MaxRAMFraction`
* Allow administrators to adjust `JVM_OPTS` and `JVM_MEMORY_OPTS` directly and separately if needed. In almost all cases, the defaults are fine, but when they aren't, you need more power than an on/off switch.
* Dependency refresh

We've already deployed similar changes to [`turbovote-admin-http-api`](https://github.com/democracyworks/turbovote-admin-http-api/pull/62) if you want to review discussion there.


eggsyntax:
Quick comment -- doesn't look like `turbovote-http-api` uses New Relic anyway (or at least not `defn-traced`), but thoughts on how to remove it from other repos? I wonder if it's worth redefining `defn-traced` in one of our own libs, currently as a no-op, and then redefining it to use datadog or whatever else we end up with.

That's probably ridiculously lazy, but it'd at least keep a lot of noise out of the repos, save some dev time, and make the switch to our new tracing really seamless.

Just tossing it out as food for thought even though it's not directly relevant to this PR.

gws:
@eggsyntax I just wrote this as a compatibility shim to do something like that: https://github.com/democracyworks/opentracing-compat/pull/1

The benefit of the OpenTracing API is that we don't need to redefine this to use a proprietary API like Datadog's, once we are using OpenTracing, Datadog's APM can hook into that, New Relic's APM can hook into it, Honeycomb.io, etc.

gws:
@eggsyntax Aside from a sweeping "update everything nobody cares about" change to drop in that new `defn-traced` macro to use OpenTracing, if you're adding OpenTracing to a service you're working on, you should use the OpenTracing Java API directly (via a Clojure wrapper), rather than that macro which is specifically for compatibility of the kind you just outlined.

eggsyntax:
> I just wrote this as a compatibility shim
Perfect :) 

> The benefit of the OpenTracing API is that we don't need to redefine this to use a proprietary API
Totally, sloppy language on my part. Really I just mean something that accomplished exactly what your shim accomplishes.

PR #44, opened 2019-09-12
Add stop ignore language

leilakass:
https://app.asana.com/0/1139475579829496/1139475579829553/f

add additional language that will cause sms notys to stop for a user or be ignored and not create zendesk tickets


src/turbovote_http_api/sms_utils.clj
```diff
@@ -34,22 +34,46 @@
...
+                           #"st0o[.!]*"
+                           #"sopt[.!]*"
+                           #"dtip[.!]*"
+                           #"remove[.!]*"
+                           #"alto[.!]*"
+                           #"leave me alone[.!]*"
+                           #"stp[.!]*"
+                           #"STOP TURBOVOTE[.!]*"
+                           #"STOP – TurboVote"
+                           #"s+t+o+p+[.!]*"})
```
mrichards42: We might think about stripping trailing punctuation as part of the standard text munging code. I'm not sure if we have situations where we care about that, and it feels excessive to have `[.!]*` on every regex.

leilakass: so we wouldn't want to strip all punctuation (by changing this https://github.com/democracyworks/turbovote-http-api/blob/7bb1e7ff78cee5d5491ed4d902a9a6ab6dcf8c7b/src/turbovote_http_api/sms_utils.clj#L13) because it may be useful for people in zendesk? 

similarly, i would think that trailing punctuation in cases that aren't a start/stop/ignore message might be helpful for zendesk to understand tone / how to respond. i could be wrong though

mrichards42: Everything in this namespace is just for matching start/stop/ignorable messages. The original message is still sent to zendesk for messages that don't match: https://github.com/democracyworks/turbovote-http-api/blob/7bb1e7ff78cee5d5491ed4d902a9a6ab6dcf8c7b/src/turbovote_http_api/handlers/sms.clj#L86-L90

leilakass: Oooh nice. So then would stripping all punctuation be cool? or is there an argument for only stripping trailing punctuation?

mrichards42: We were trying to be conservative about not matching too many messages -- I'd be curious to hear some input from @eggsyntax about that tradeoff if you remember any more about it.

My inclination is that stripping all punctuation would be fine. I was suggesting just trailing punctuation since that's a more conservative change, which would be less about changing the logic, and more about reducing some repetition in these regexes.

leilakass: Yeah I think stripping all the punctuation would be good because I can't really see when that would change the meaning such that we wouldn't want to ignore a message. And it is likely to come up when the reply is two words like "leave! me! alone!" lol

eggsyntax: We did definitely want to err on the conservative side -- I suggest either
- going with the more conservative option, or
- checking with Esha/Carmen and Ada to make sure they don't think the less conservative option will cause problems -- Ada in particular has seen about ten zillion of these responses and would probably have a good sense of whether there's likely to be a problem.

src/turbovote_http_api/sms_utils.clj
```diff
@@ -34,22 +34,46 @@
 ;;; Stop messages
 
 ;; per https://support.twilio.com/hc/en-us/articles/223182208-Industry-standards-for-U-S-short-code-HELP-and-STOP
-(def standard-stop-words #{"stop" "cancel" "unsubscribe" "quit" "end"})
+(def standard-stop-words #{"stop" "stopall" "cancel" "unsubscribe" "quit" "end"})
```
eggsyntax: The intent behind the separate `standard-stop-words` is that they're the ones that meet a specific standard (see the "STOP message" section of the linked document, and as a result they're the ones that match Twilio's automatic filtering (which is what we were using until we switched to sending from a shortcode). So unless there's a change to that standard, I think `stopall` should be in `extended-stop-words`.

leilakass: I'm definitely going crazy because both @mrichards42  and I saw that 'stopall' was a standard word but now when I look back it's not there...... will remove it again lol https://support.twilio.com/hc/en-us/articles/223182208-Industry-standards-for-U-S-short-code-HELP-and-STOP

mrichards42: It's not at that link, it's at https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-

mrichards42: Though it does seem like we might be able to remove all the default replies using the "Advanced Opt-Out for Messaging Services" settings from that link.

leilakass: Would we want to remove all the default keywords or add our additional keywords into Twilio?

mrichards42: Not sure -- either one would make this code simpler. Probably depends on how flexible their keyword matching is, and how easy it is to customize the response.

mrichards42: But also I'm not sure I'd bother with that right now since it might be a rabbit hole (and someone from product can also look into it)

leilakass: I looked a bit into it and there is a maximum of 20 opt-out keywords. I think it would be better to handle it on our end because there's no limit on keywords and we can also choose to exclude punctuation (which Carmen + Esha said they are interested in).

eggsyntax: Oh, interesting, I think the Advanced Opt-Out is new since last year -- I don't think they offered opt-out keywords for shortcodes at all then. The biggest advantage of doing it ourselves is the flexibility, but if it's flexible enough, I'm always in favor of deleting code ;)

eggsyntax: Oh, just saw your latest response. Yeah, I don't think 20 is enough for us, especially because there are (or were) at least vague plans of making our handling more sophisticated.

leilakass: I won't remove the default keywords or change anything about this right now though

eggsyntax: As far as I know, Zendesk folks still see lots of messages from users that want to be unsubscribed that these filters don't catch, so I think there's a *lot* of room to improve the automated filtering and decrease the burden on human attention.

src/turbovote_http_api/sms_utils.clj
```diff
@@ -34,28 +34,51 @@
...
 
 ;; per Esha's analysis of common stop messages in zendesk tickets. Should not be
 ;; replaced by an edit distance approach due to high risk of false positives (and
 ;; low cost of false negatives, ie it just goes to zendesk).
-(def extended-stop-words #{"stoo" "stip" "sto" "stol" "top" "s top" "s t o p"
-                           "opt out" "srop" "stio" "ztop" "ALTO – TurboVote"
-                           "Syop" "Stpp" "atop"  "sto0" "st0o" "sopt" "dtip"
-                           "remove" "alto"})
+(def extended-stop-words #{#"stoo+"
```
eggsyntax: `#"stoo+"` matches `"stoo", "stooo", "stoooo", "stoooo,", ...` and nothing else -- was that what you intended?

leilakass: yeah, I thought it would be helpful because people tend to repeat the last character of their text. like 'stoppppppppp' and would be helpful to catch that even if they misspell the word. an idea, not opposed to changing it back.

eggsyntax: Cool, I thought it might be that, but I wasn't sure if that had gotten common. I don't see any likely bad outcomes, so no preference either way for me. If you stick with it, it looks like a couple can be merged, eg `sto+` covers `stoo+`, but that's incredibly minor...

src/turbovote_http_api/sms_utils.clj
```diff
@@ -20,12 +20,18 @@
   [s]
   (clojure.string/replace s really-emo-regex ""))
 
+(defn normalize-whitespace
+      "Use regex to replace multiple whitespaces with a single white space."
+      [s]
+      (clojure.string/replace s #" +" " "))
```
mrichards42: nit: indentiation

PR #45, opened 2019-09-18
Update Jetty max-threads to 100

gws:
This allows us to make more out of the HTTP API, since we are giving it 4GiB of RAM in production. Since this service is spending the vast (vast) majority of its time waiting for IO, this number could easily be further adjusted upward, however, the more RabbitMQ producers we have available the better throughput we can get if specific channels become busy.

This parameter is also going to be necessarily tied to the HPA scaling targets, where we will scale out when more than 70 threads (70% of capacity) are in use averaged across the fleet.


gws:
Prior to this change, the default max-threads was used by Jetty, which was **50**. That is what we went to production with in 2018 AFAIK.

gws:
For reference: https://github.com/democracyworks/kubernetes-scaling/commit/6fe82143e7992c3d33a2e34865e4a10839c4e877

That might not be in anyone's wheelhouse to review but wanted to link these two changes here if anyone else stumbles across this PR in the future.

src/turbovote_http_api/service.clj
```diff
@@ -104,8 +104,10 @@
                                      (map re-pattern
                                           (config [:server :allowed-origins])))
                                     :max-age (* 5 60)}
-       ::bootstrap/host (or (config [:server :hostname] "localhost") "localhost")
        ::bootstrap/type :jetty
+       ;; Jetty-specific options: http://pedestal.io/reference/jetty
+       ::bootstrap/container-options {:max-threads 100}
```
mrichards42: Should that be `(or (config [:server :max-threads] 100) 100)`? Totally fine if you want to hard-code it and not worry about it right now.

gws: I'm considering just leaving it, the 50 max threads was hardcoded already, and changing the config requires a rollout anyway (it doesn't hurt to have extra available threads lying around in dev / staging either, if it did I think I'd want it configurable, though).

PR #46, opened 2019-11-12
Fix CORS for deploy previews

mrichards42:
Also removes dev CORS config from production deploys.


deploy/development/turbovote-http-api.configmap.yml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: turbovote-http-api
+data:
+  ALLOWED_ORIGINS: '["https://(.+[.])?turbovote[.]org" "https://(.+[.])?dev[.]democracy[.]works" "https://deploy-preview-[0-9]+--turbovote-cdn.netlify.com"]'
```
gws: These deploy preview URLs come in a lot of flavors, over on Ballot Scout they use `"https://.+--ballotscout.netlify.com"`, is there a reason that'd be too permissive?

mrichards42: Good point. We usually only using the `deploy-preview-NNN--` version, but it's possible to get to the page from a variety of urls, so it would probably be safer to make the whole prefix a wildcard.

PR #47, opened 2020-01-17
Set up endpoint for enriching zendesk tickets

dzcode:
Sibling zendesk-works PR https://github.com/democracyworks/zendesk-works/pull/20
Ticket: https://test-ve.atlassian.net/browse/VE-91


src/turbovote_http_api/handlers/zendesk.clj
```diff
@@ -51,10 +56,23 @@
...
+  [["/"
+    ^:interceptors [(negotiate-response-content-type ["text/plain"])
+                    authorize-zendesk
+                    zendesk-response]
+    {:post [::post-zendesk
+            (bifrost/interceptor
+             incoming (utils/timeout-config [:zendesk :incoming]))]}]
+
+   ["/enrich"
+    ^:interceptors [(negotiate-response-content-type ["text/plain"])
```
dzcode: I'm not sure if the `(negotiate-response-content-type ["text/plain"])` and `zendesk-response` interceptors are strictly necessary here

mrichards42: Since this is being called from Zendesk, we just need to return a response that Zendesk can handle right? I'm not sure what response format they're looking for.

src/turbovote_http_api/handlers/zendesk.clj
```diff
@@ -51,10 +56,23 @@
     (fn [ctx]
       (assoc ctx :response (ring-resp/response "OK")))}))
 
+(defn enrich-ticket
+  [ctx]
+  (enrich-ticket-fire-and-forget (some-> ctx :request :body-params))
```
mrichards42: Is it OK to call `enrich-ticket-fire-and-forget` with `nil`? I'm assuming somewhere we put the argument on an async channel.

dzcode: That's a good point, I'm not sure, but it'd be easier and make more sense to just throw an error here, so I'll do that

src/turbovote_http_api/handlers/zendesk.clj
dzcode: I was following the pattern above which is also called from Zendesk, and just responds with a plaintext "OK"

dzcode:
Updated the description with:
Sibling zendesk-works PR democracyworks/zendesk-works#20
Ticket: https://test-ve.atlassian.net/browse/VE-91

src/turbovote_http_api/handlers/zendesk.clj
mrichards42: Just a minor thing here -- I believe throwing an exception will result in a server error response, but if the problem is that the request is missing a body, that's probably a 400.

src/turbovote_http_api/handlers/zendesk.clj
mrichards42: How is Zendesk calling this endpoint? Is it channelback, like `/`, or a webhook, or something else?

dzcode: It'll be a webhook sending a post request with authz headers

mrichards42: Gotcha -- I couldn't find much info about the response format Zendesk expects w/ webhooks, so if it works when you test it 🤷‍♂. FWIW, if the existing endpoint is actually channelback, it looks like we're supposed to [return json](https://developer.zendesk.com/apps/docs/channels-framework/channelback#response-format), but I assume that's been working so who knows.

dzcode: Actually looking further, it'll be an HTTP target, if that's any different. That appears to be how the zendesk SMS responses are being handled as well from the zendesk end

mrichards42: Oh, if SMS responses are actually an HTTP target webhook, we should update the comment on `incoming` since it says it's channelback.

src/turbovote_http_api/handlers/zendesk.clj
```diff
@@ -49,12 +54,32 @@
   (pedestal/interceptor
    {:leave
     (fn [ctx]
-      (assoc ctx :response (ring-resp/response "OK")))}))
+      (if (:response ctx)
+        ctx
+        (assoc ctx :response (ring-resp/response "OK"))))}))
```
mrichards42: Pedestal [short-circuits](http://pedestal.io/reference/servlet-interceptor#_early_termination) as soon as there is a `:response` on the context, so you don't have to worry about this change. In fact we rely on that behavior in the `authorize-zendesk` interceptor, which adds the 403 response (before the `zendesk-response` handler has a chance to process the request).

src/turbovote_http_api/handlers/zendesk.clj
```diff
@@ -49,12 +54,32 @@
...
+
+(defn enrich-ticket
+  [ctx]
+  (if-let [params (some-> ctx :request :body-params)]
+    (do
+      (enrich-ticket-fire-and-forget params)
+      ctx)
+    (assoc ctx :response {:status 400,
+                          :headers {}
+                          :body  "No body params in request"})))
```
mrichards42: This is totally fine, but slight preference to use either `utils/assoc-response` or [`ring-response/bad-request`](https://github.com/ring-clojure/ring/blob/7a87a3f8d52285e144d51c0d18602c5294f08a9d/ring-core/src/ring/util/response.clj#L47) to avoid a little boilerplate and the potential for misspellings.

src/turbovote_http_api/handlers/zendesk.clj
dzcode: In this case, that doesn't wind up applying since the `:leave` part of `zendesk-response` handler is called upon short-circuiting (from the docs, it looks like any of the leave functions are called from interceptors that were already invoked).

With that said, I think this small refactor winds up being cleaner anyway

mrichards42: Oh yup, you're right. I missed that `enrich-ticket` was last in the chain. Agreed that what you ended up with is better in any case.

mrichards42: I suppose the other option, which is the route we've taken in some other places, is to isolate the action in the final interceptor, and do all the validation in separate interceptors, so, e.g.

```clj
[authorize-zendesk
 validate-zendesk-request ;; <-- the new one
 zendesk-response
 enrich-ticket]
```

But in this pretty simple case I wouldn't say that's necessary. If we had a bunch of zendesk handlers that all needed the same validation it might be worth it, but we aren't there yet.

PR #48, opened 2020-01-31
Allow http server port to be configured

mrichards42:
... but still defaults to 8080


mrichards42:
@sean that's actually not a double-default, it's an annoyance w/ how `resource-config` handles missing values (an exception):

```clj
;; when some value is missing
(config [:some :value])              ; => exception
(config [:some :value] 123)          ; => 123
;; when some value is present but nil
(config [:some :value])              ; => nil
(config [:some :value] 123)          ; => nil
(or (config [:some :value] 123) 123) ; => 123

```

sduckett:
> it's an annoyance w/ how `resource-config` handles missing values (an exception):
yeah, that's what I was afraid of.

PR #49, opened 2020-02-03
Handle kehaar timeouts

mrichards42:
Using the new `bifrost.core/handler` from democracyworks/bifrost#7


src/turbovote_http_api/service.clj
```diff
@@ -57,9 +58,20 @@
...
                  ctx)))}))
 
+(def log-http-errors
+  (pedestal/interceptor
+   {:name ::log-http-errors
+    :leave (fn [ctx]
+             (let [status (get-in ctx [:response :status])]
+               (when (and (int? status) (<= 400 status))
+                 (log/warn "Sending error response"
+                           (pr-str (:response ctx)))))
```
mrichards42: Not sure if we want to log additional info (the request?), or something else entirely here (or if we want to rely on something other than logging).

src/turbovote_http_api/service.clj
```diff
@@ -57,9 +58,20 @@
                      (assoc-in [:request :headers "content-type"] "application/json"))
                  ctx)))}))
 
+(def log-http-errors
+  (pedestal/interceptor
+   {:name ::log-http-errors
+    :leave (fn [ctx]
+             (let [status (get-in ctx [:response :status])]
+               (when (and (int? status) (<= 400 status))
```
gws: I'd lean toward just logging 5xx here since 4xx are best resolved at the client, and would consider using the `error` level since it's something we aim to eliminate from the system.

You don't have to do this! Just a thought.

src/turbovote_http_api/service.clj
gws: Yeah, I'd like to log less PII and avoid logging things like `Authorization` headers but I also like logging requests conceptually (for 5xx errors), I don't know how often requests have actually been useful in diagnosing the cause of a 5xx (usually, here, it's a downstream service that has better information about what happened), so it might depend more on how often you'd like to have had the request here in the past...

mrichards42: Agreed RE: excluding PII -- the one thing that I think would be especially helpful that I've wanted in the past is the http verb and path so it's clear which backend service to point the finger at.

src/turbovote_http_api/service.clj
mrichards42: Agreed in general. I might leave the 400s in here for now and take them out if they become noisy, since by and large the only client that uses this api is turbovote itself, so it might reveal errors we've been unable to capture in tv-web.

src/turbovote_http_api/service.clj
mrichards42: I added what I believe are the important request keys (method, server, uri, query string) and left out everything else, which I think should cover PII concerns.

PR #50, opened 2020-02-03
Handle JWTs with missing information

mrichards42:
These should return the same response as an invalid JWT (403).

Working on getting rid of junk errors ⚡️


src/turbovote_http_api/handlers/election_notifications.clj
```diff
@@ -37,12 +37,16 @@
...
     (fn [ctx]
       (let [unsigned-token (utils/validate-token
                             (get-in ctx [:request :path-params :user-token]))
-            user-id (java.util.UUID/fromString (:user-id unsigned-token))]
-        (if (not= (::utils/status unsigned-token) :error)
+            user-id (try
+                      (java.util.UUID/fromString (:user-id unsigned-token))
+                      (catch IllegalArgumentException _
+                        nil))]
+        (if (and user-id (not= (::utils/status unsigned-token) :error))
```
dzcode: Why would a missing user-id be a 403 and not a 400?

mrichards42: b/c this is supposed to be an endpoint that requires some sort of authorization

src/turbovote_http_api/handlers/election_notifications.clj
```diff
@@ -37,17 +37,21 @@
...
+                              java.util.UUID/fromString)
+                      (catch IllegalArgumentException _
+                        nil))]
+        (if (and user-id (not= (::utils/status unsigned-token) :error))
           (-> (assoc-in ctx [:request :path-params :user-id] user-id)
               (assoc-in  [:request :path-params :mediums] #{:email})
               (update-in [:request :path-params] dissoc :user-token))
-          (assoc-in ctx [:response :status]
-                    403))))}))
+          (utils/assoc-response ctx 403 "Not authorized"))))}))
```
mrichards42: A response with just a `:status` key doesn't actually do anything -- it has to be an actual ring response. So this interceptor was actually just letting the request pass through unchanged when it was invalid.

src/turbovote_http_api/handlers/election_notifications.clj
```diff
@@ -37,17 +37,21 @@
...
+   {:name ::validate-token
+    :enter
     (fn [ctx]
       (let [unsigned-token (utils/validate-token
                             (get-in ctx [:request :path-params :user-token]))
-            user-id (java.util.UUID/fromString (:user-id unsigned-token))]
-        (if (not= (::utils/status unsigned-token) :error)
+            user-id (try
+                      (some-> (:user-id unsigned-token)
+                              java.util.UUID/fromString)
```
mrichards42: `UUID/fromString` actually throws `NullPointerException` when given `nil`, so safeguarding against this.

src/turbovote_http_api/handlers/election_notifications.clj
dzcode: Looks like `utils/assoc-unauthorized` does this line

mrichards42: 👍 that sounds like a good refactor to me -- done

PR #52, opened 2020-03-26
[wip] [skip ci] initial commit for draft

solomonsokoya:
Running into a 504 gateway problem when I try to communicate to this new endpoint. Any suggestions? 

Is it because I have not made changes to user-works?


src/turbovote_http_api/service.clj
```diff
@@ -77,8 +78,15 @@
                            (pr-str (ctx-for-log ctx)))))
              ctx)}))
 
-(def routes
-  [[(-> ["/"
+(def tracking-app
+  [:tracking-app
+   (-> ["/"
+        ^:interceptors [;;TODO interceptors like main-api??]]
```
mrichards42: yeah, I'm guessing we'll want most of the same interceptors, so probably best to factor those out into a separate `def`, and then take out the ones that aren't shared between the two apps.

src/turbovote_http_api/handlers/utils.clj
```diff
@@ -55,6 +55,12 @@
                                 #(UUID/fromString %))
    :name ::user-id-from-path))
 
+(def event-timing-from-path
```
mrichards42: This is super specific to the new timing api. I'd go ahead and put it in the event-timing namespace.

src/turbovote_http_api/service.clj
```diff
@@ -77,8 +78,15 @@
...
+       (utils/mount-routes "/event-timing" event-timing/routes))])
```
mrichards42: This route is different than what we set up over in the frontend. We should pick one or the other.

src/turbovote_http_api/handlers/user_ready_timing.clj
```diff
@@ -0,0 +1,40 @@
...
+  (assoc
+   (bifrost.i/update-in-request [:query-params :event-timing]
+                                #(keyword %))
+   :name :event-timing-from-path))
+
+(def user-id-from-path
+  (assoc
+   (bifrost.i/update-in-request [:query-params :id]
+                                #(UUID/fromString %))
+   :name ::user-id-from-path))
```
mrichards42, line 32: In the context of a request, "path" is the part of the url without the query string, so these functions should probably be called `*-from-query` . . . then again, most of our api uses the url and not query params (e.g. [/election-notifications/subscriptions/:user-id](https://github.com/democracyworks/turbovote-http-api/blob/155b72922c9b834b7cb291d177918853e93cfd92/src/turbovote_http_api/handlers/election_notifications.clj#L61) and [/voting-methods/preference/:user-id](https://github.com/democracyworks/turbovote-http-api/blob/e75263394e60fea52892fb00697c311dd656a077/src/turbovote_http_api/handlers/voting_methods.clj#L27)) so it would probably be more consistent with the rest of our api to at least put the user-id part of this in the path, i.e. `/:user-id/:event`, or `/:user-id?event=:event`

src/turbovote_http_api/handlers/user_ready_timing.clj
```diff
@@ -0,0 +1,40 @@
...
+
+(def routes
+   ["/"
+    ^:interceptors[action-from-path
+                   user-id-from-path]
+    {:post [::track-event
+            (bifrost/handler update
+                             (utils/timeout-config [:user :update]))]}])
```
mrichards42, line 40: I know we don't actually have any bifrost timeouts in our config (which is what this is; we do have plenty of _kehaar_ timeouts in the config, just to make things confusing) . . . but we probably want to change this path to be something specific to this route, instead of using the user-update timeout.

solomonsokoya:
I did not go forward with the test because the [fire-and-forget](https://github.com/democracyworks/turbovote-http-api/blob/18fbf77c740e8d46290bfaa360d29c2efe35c5d0/test/turbovote_http_api/test_utils.clj#L54) function with this particular change doesn't return nothing back.

src/turbovote_http_api/handlers/user_ready_timing.clj
```diff
@@ -0,0 +1,36 @@
...
+(def update (wire-up/async->fn update-channel))
+
+(defn close-all-channels! []
+  (async/close! update-channel))
+
+(def action-from-query
+  (assoc
+   (bifrost.i/update-in-request [:query-params :event-timing]
+                                #(keyword %))
+   :name :event-timing-from-path))
```
sduckett, line 24: Echoing [a previous comment](https://github.com/democracyworks/turbovote-http-api/pull/52#discussion_r401916850): Would it feel consistent to have `action-from-query` put the thing in `:event-timing-from-query` (since we're looking at a query-param) instead of `event-timing-from-path`?

solomonsokoya: whoops. will changes

resources/config.edn
```diff
@@ -100,6 +100,10 @@
                        :channel turbovote-http-api.handlers.users/update-channel
                        :response true
                        :timeout 12000}
+                      {:queue "user-works.user.update-user-ready-timing"
+                       :channel turbovote-http-api.handlers.user-ready-timing/update-channel
+                       :response true
+                       :timeout 12000}
```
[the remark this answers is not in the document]
mrichards42: Definitely just a suggestion though -- what you have should work fine solomon, if you'd prefer to keep it how it is.

src/turbovote_http_api/handlers/user_ready_timing.clj
```diff
@@ -0,0 +1,36 @@
...
+
+(def action-from-query
+  (assoc
+   (bifrost.i/update-in-request [:query-params :event-timing]
+                                #(keyword %))
+   :name :event-timing-from-query))
+
+(def routes
+  ["/"
+   ["/:user-id"
```
mrichards42, line 28: This would be a great place to add a constraint, especially if `event-timing` is going to be a query param.

```clj
;; I think this is the right syntax, not totally sure. What this should do is
;; look for `:event-timing` as a param (in :query-params, :path-params, or
;; :body-params) and validate it using a function (in this case, the set of
;; allowed values). If it doesn't match one of those values, the whole route
;; doesn't match, so you get a 404
["/:user-id"
 ^:constraints {:event-timing #{"done" "not-done" "drop-off"}}
 ^:interceptors [...]
 {:post [...]}]
```

Here are a couple examples from the election routes: https://github.com/democracyworks/turbovote-http-api/blob/14b7605504459c2d74bafad2da3dabb046401ffe/src/turbovote_http_api/handlers/elections.clj#L29-L48

resources/config.edn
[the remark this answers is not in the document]
sduckett: I'm on team kehaar events, @mrichards42: It's (probably?) the ideal match for the idea we're talking about here. It might even be easier to test?

src/turbovote_http_api/handlers/signup_event.clj
```diff
@@ -0,0 +1,43 @@
...
+(def action-from-query
+  (assoc
+   (bifrost.i/update-in-request [:query-params :action]
+                                #(keyword %))
+   :name :action-from-query))
+
+(def routes
+  ["/"
+   ["/:user-id"
+    ^:constraints {:action #"done|not-done|drop-off"}
```
mrichards42, line 39: Does this work? I'm not entirely up on the syntax, but I assumed constraints need to be functions, or things that can act like functions (e.g. sets, maps, but not regexes).

If so 👍, otherwise I think you can make this a set of acceptable strings:
```clj
{:action #{"done" "not-done" "drop-off"}}
```

You should be able to test this by sending some requests and seeing what kind of response you get back:

```sh
# known action: expect 200
curl -i -XPOST "http://localhost:$PORT/signup-event/$USER_ID?action=done"

# unknown action: expect 404
curl -i -XPOST "http://localhost:$PORT/signup-event/$USER_ID?action=fake"

# no action: expect 404
curl -i -XPOST "http://localhost:$PORT/signup-event/$USER_ID"
```

src/turbovote_http_api/handlers/signup_event.clj
```diff
@@ -0,0 +1,43 @@
...
+             (utils/assoc-response ctx 200 "OK"))}))
+
+(defn close-all-channels! []
+  (async/close! signup-event-channel))
+
+(def action-from-query
+  (assoc
+   (bifrost.i/update-in-request [:query-params :action]
+                                #(keyword %))
+   :name :action-from-query))
```
mrichards42, line 34: Nit: I'd used a namespaced keyword for this (i.e. instead of `:action-from-query`, use the double-colon `::action-from-query`, which expands to `:turbovote-http-api.handlers.signup-event/action-from-query`). If we ever need to inspect this interceptor during a request, it'll be helpful to have the namespace as part of it.

src/turbovote_http_api/service.clj
```diff
@@ -77,8 +78,17 @@
...
              ctx)}))
 
-(def routes
-  [[(-> ["/"
+(def signup-event-app
+  [:signup-event-app
+   (-> ["/"
+        ^:interceptors [log-http-errors
+                        (negotiate-response-content-type
+                         ["text/plain"])]]
```
mrichards42, line 86: It might be helpful for future readers if you added a comment about how we are specifically _not_ including the `(body-params)` interceptor, which is one of the standard interceptors. IIRC we're excluding it b/c we expect to get POST requests with `Content-type: text/plain`, and (a) the `body-params` interceptor doesn't know how to handle text/plain, and (b) we don't expect to get any content, so it's safe to ignore any body params.

src/turbovote_http_api/handlers/signup_event.clj
solomonsokoya: It works. I tried using a map but I got these errors

mrichards42: Huh, you're totally right, it expects a regex, not a function: https://github.com/pedestal/pedestal/blob/3ba98b88b93f1781b9cdbce0670993e9e8856cdf/route/src/io/pedestal/http/route/linear_search.clj#L32-L35

Ignore me 🙂

test/turbovote_http_api/signup_event_test.clj
```diff
@@ -0,0 +1,34 @@
...
+             [clojure.string :as str])
+  (:import (java.util UUID)))
+
+(def root-url "/signup-event")
+
+(defn structure-url [action id]
+  (str root-url
+       "/"
+       id
+       "/?action="
```
sduckett, line 16: ❓ Related to my comment on https://github.com/democracyworks/turbovote-web/pull/361, why a slash after the UUID and before the query params?

PR #54, opened 2020-04-06
Support netlify.app in development

gws:
This relates to a [change](https://community.netlify.com/t/changes-coming-to-netlify-site-urls/8918) Netlify will be making this month.

An identical change has been tested and rolled out on Ballot Scout. This does not need to be released to production as it only impacts development.


gws:
@sduckett I don't think you read it _wrong_, but the change in domain does have implications for CORS that they don't call out in the announcement. I don't know if they didn't think it through or if CORS on non-custom domains is just rare.

There's also this, emphasis mine:

>Do I need to update my site or application?
>
>**In most cases, no**. This change is very unlikely to impact the way your application functions, as all requests will automatically be forwarded to netlify.app on your behalf.
What about new sites?

PR #55, opened 2020-04-10
Update README

solomonsokoya:
Give a lil context to adding routes.


README.md
```diff
@@ -95,6 +95,9 @@ In order to run `turbovote-http-api` in dev connected to krakenstein, you need t
                          :port 45672} ; krakenstein
 ```
 
+## Adding Routes
+This service uses [Pedestal](http://pedestal.io/guides/defining-routes) to build our API. Depending on the purpose of the route you're adding you can add a new "app" to the routing table OR within the `main-api-app` you can use the given helper `utils/mount-routes` to add a new route. 
+
```
eggsyntax, line 100: Psyched to have some more documentation here! At the risk of being pedantic, a couple of things that I think would help:
- what exactly does "app" mean here?
- how do I choose between the two options you describe?
- examples of the two would be handy but NBD, and/or links to the right spot in the code.

All that said, having this as-is is better than not having it, so don't let any of that stop you from merging! :grinning:

